PR #20, opened 2015-04-14
Use dynamic locale and scan for the right locale file to use.

andreas-eberle:


justangel:
works perfect, thank you

PR #23, opened 2015-04-15
Adjust the height which needs to be rendered additionally to the bottom depending on scale. Fixes #15.

### LINKED ISSUE #15: Terrain Frustum problem.
I notice terrain becomes black at the bottom part of the screen when moving over mountains, this does not happen on flat ground level terrain, it looks like frustum culling problem to me, as if the tiles are outside frustum while they should be inside.
(this happens in both 1.0 and 1.1 release builds on any map)

andreas-eberle:
Please test.


andreas-eberle:
As this fix is already in the master (see e8c8c96813910b44e1828adcd781a455192f29f0), I close this request.

PR #28, opened 2015-04-16
Mechanism added to avoid main menu null item actions

adamjryan:
Update to the main menu / sub menus to disable the submit
button when there is no item selected to prevent processing
a null object.

### DIFF
--- a/jsettlers.graphics/src/jsettlers/graphics/utils/UIList.java
+++ b/jsettlers.graphics/src/jsettlers/graphics/utils/UIList.java
@@ -49,20 +49,22 @@ public class UIList<T> implements UIElement {
 		UIListItem getItem(T item);
 	}
 
-	public UIList(List<? extends T> items, ListItemGenerator<T> generator,
-			float itemheight) {
-		this.items = items;
+	public UIList(List<? extends T> items, ListItemGenerator<T> generator, float itemheight) {
+		setItems(items);
 		this.generator = generator;
 		this.itemheight = itemheight;
-		if (items.size() > 0) {
-			activeItem = items.get(0);
-		}
 	}
 
 	public void setItems(List<? extends T> list) {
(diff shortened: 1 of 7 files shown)


michaelzangl:
Looks good so far. I will test it (and probably merge it) this weekend. Thanks ;-)

andreas-eberle:
@michaelzangl: Any progress here?

adamjryan:
Let me know if there are any changes needed for this, so it can get sorted.

jsettlers.graphics/src/jsettlers/graphics/utils/UIList.java
```diff
@@ -49,19 +49,18 @@
...
 		this.itemheight = itemheight;
-		if (items.size() > 0) {
-			activeItem = items.get(0);
-		}
 	}
 
 	public void setItems(List<? extends T> list) {
 		synchronized (itemsMutex) {
 			this.items = list;
+			if (items.size() > 0) {
```
michaelzangl: This is not the way it was intended. Lists are often updated by adding items. If we reset the active item on every list update, we reset the user selection.

adamjryan: If a new set of items is set for the UIList then the previous active item would be invalid. In fact when the new list size is 0 the active item should be set to null. There should also really be a check on the setActiveItem method as well to check it's in the list.
Additions would be made on the List object (that's passed in) and these will be reflected in the UI without changing the user's selection.

adamjryan: I think maybe you meant that you want the previous selection to be kept when going back to a menu. I've updated it now so the active item for each menu is not reset when clicking between menus.

michaelzangl:
Merged in controls-improve-menu.

PR #43, opened 2015-04-20
Introduce dialog to question the user for his Settlers III original installation folder (fixes #8)

### LINKED ISSUE #8: The folder of the original Settlers installation must be requested on first startup
Instead of letting the user configure the folder in the config file, we should open a dialog on first startup or whenever the settlers files cannot be found. This dialog should allow you to select the folder and save this to the configuration file.

andreas-eberle:
This implements the changes for #8 and changes the handling of the config.prp file. The default settings are now stored in the jar file, whereas only the local settings are stored in the config.prp file. This is better when you update the game, as the settings do not need to be overwritten.

@michaelzangl: Can you have a look on my changes? It seems to work properly with the Swing UI, but the Android UI misses the images in the building menu when you try to construct a building. 

The weird thing is that it seems to me that all the required graphics have been generated and packaged into the apk file. Do you have an idea?

### DIFF
--- a/jsettlers.graphics.androidui/build/jsettlers/graphics/androidui/AndroidUiImageGenerator.java
+++ b/jsettlers.graphics.androidui/build/jsettlers/graphics/androidui/AndroidUiImageGenerator.java
@@ -61,7 +61,7 @@ public class AndroidUiImageGenerator {
 		if (configFile == null) {
 			throw new RuntimeException("please use configFile=\"...\"");
 		}
-System.err.println(configFile);
+
 		SwingResourceLoader.setupGraphicsAndSoundResources(new ConfigurationPropertiesFile(configFile));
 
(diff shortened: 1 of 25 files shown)


jsettlers.graphics.androidui/build/jsettlers/graphics/androidui/AndroidUiImageGenerator.java
```diff
@@ -60,8 +61,8 @@ public void execute() throws IOException {
 		if (configFile == null) {
 			throw new RuntimeException("please use configFile=\"...\"");
 		}
-
-		SwingResourceLoader.setupSwingResourcesByConfigFile(configFile);
+System.err.println(configFile);
```
michaelzangl: TODO: Remove this

andreas-eberle:
@michaelzangl: I was able to find and fix the bug.

Issue #53, opened 2015-04-25
If automatic locale search fails show english as default instead nothing


Fantu:
I did some test of build from master, I saw that automatically search locales, it loaded partial italian I did but don't show english for strings missed (for this probably I can found a workaround with transifex)
Another bigger problem if when not found rispective locales where don't show all strings, I think is good use english as default if not found locale.

andreas-eberle:
@Fantu: your right. I reopen the ticket. My fault.

Issue #50, opened 2015-04-23
small string error in english


Fantu:
I found a small string error in english:
big tmeple
that should be:
big temple

there are also other strings strange probably to improve

about translation project in trasifex probably can be good add a refer about it somewhere:
https://www.transifex.com/projects/p/settlers3-remake/
anyone can use it and should make easier and faster translation in any languages, from web, without programming experience, also with more people working on same translation in real time ecc...

andreas-eberle:
Hi Fantu,

how is the work with transifex exactly? I joined your project there to translate our settlers 3 remake and was already able to change some buggy labels. But how are the changes propagated back into this repository? 

Furthermore, do you, as project maintainer, have an easy way to download all the labels files with their proper naming? I was only able to download the files one by one and with other filenames as in the repository. 

Another aspect: Is it possible to add a required label to all the translation files at once? I mean, if we have some new labels, they are needed in all the translation files. Do I have to copy them to all files, or is there a functionality to do this automatically. 

I think if there are solutions for these questions, we could really work with transifex.

Best regards + many thanks for your proposal,
Andreas

Fantu:
Thanks for your reply.
About transifex:
I can't do same name in transifex for resources and mapcreator files because have same name (label_<language>.properties)
For now download/upload is manual but is possible automatize it with client and scripts with api.
For example this for another project:
https://bitbucket.org/Swyter/warband-acok-l10n/src
with settlers should be easier because conversion is not needed, transifex had already java properties files support.
I'll probably do at least small bash script in weekend probably.
A problem for automatic scripts can be wrong language code, for example "dk_DK" should be "da_DK" 
About russian will be the only and changed to "ru" or I must change it on transifex to "ru_RU" ?

When template language is updated all, transifex automatically update all languages, maintaining translations and adding/changing only lines added and changed, with acok in which each new version changes even thousands of strings transifex has saved me many hours of work.

About string error... "big tmeple" is still wrong

Fantu:
another small writing mistake:
"big living house," should be "big living house"

PR #57, opened 2015-04-26
Fixed small locales error

Fantu:
Fixed the default language to en if user language file is missed.
Fixed 2 small type errors in english locale file.


Fantu:
this partially fixes #53 fix only user locale file missed but not strings missed in locale selected
also partially fixes #50 with 2 small type error found but probably other strings needs fixes and/or improvements

PR #63, opened 2015-04-29
remove all duplicates from movables.txt

justangel:


justangel:
maybe I'd better split into several commits? To avoid confusion?

michaelzangl:
No, you don't need to ;-). I just need some time to look over it.

Issue #58, opened 2015-04-27
Danish locale naming dk_DK instead of da_DK


andreas-eberle:
@stumpdk: You named the danish locale file "labels_dk_DK.properties". However, in the list of [supported locales of JAVA 8](http://www.oracle.com/technetwork/java/javase/java8locales-2095355.html), the danish locale is specified as da_DK. 

What is the exact locale of your system? Would it work for you to change the file nameing to "labels_da_DK.properties"?

Fantu:
Or probably is better only "da" for support any danish in each country they are...

PR #64, opened 2015-04-30
Rename labels_dk_DK.properties to labels_da_DK.properties

stumpdk:
To comply to the localization standards.


stumpdk:
Fixes [#58](https://github.com/jsettlers/settlers-remake/issues/58)

Fantu:
I think that will be better "da" instead to have it automatically working for all danish in each country and not only in denmark

stumpdk:
Da is only used in one country, DK, according to the Java localization standard.

PR #65, opened 2015-05-01
Worker drop animation correction.

adamjryan:
Previously for some workers after dropping off their newly
crafted material they would still be holding it when returning
to the standing position. This fix ensures that after it has been
dropped the material is removed from their person.


adamjryan:
I've noticed this is actually less of an issue with changes to the scripts as they don't stand around as long after dropping the material. It would be noticeable again if the timings change in future.

andreas-eberle:
@adamjryan: After looking into this with @michaelzangl, we think there is a bigger change needed. I described it in #72. With that change this will become obsolete. So I'll close this pull request without merging it.

adamjryan:
Ok, thanks for letting me know.

PR #66, opened 2015-05-02
Visual update for the in-game control panel.

adamjryan:
# The sidebar frame now has an adaptive response to use higher
resolution graphics when the window size allows.
# Mini-map buttons added.
# The mini-map frame has been adjusted to overlap the mini-map
and better fit the shape.
# The primary tabs now contain all the correct secondary tab
buttons.
# The button placeholders visible at the bottom of the
sidebar panel are now overlaid with button components
using the corresponding graphic.


adamjryan:
hmm won't be using "#" in my commit mesage next time

PR #82, opened 2015-05-15
fixed images on the left when bearer is selected

justangel:


michaelzangl:
I guess you are using the demo (or the amazon Demo). I know that some image numbers are changed in the Demo and the images you changed were one of them. They seem right for me (using Settlers III Gold files). Can you please state your version? We can start adding version guards (if/else), but I want it as documented as possible. The best thing would be to find the pattern behind the changes (like in image file xxx, add 6 for every image above 100). That way, we could simply translate the demo images.

justangel:
i'm using S3 Gold edition v.1.60

justangel:
in my version i have all addons, full version of the game

justangel:
i have two CD of Gold Edotion, 700+MB  each, with audio tracks
Soon I reinstall the game to determine the priority and the differences between the image files
since I'm sure I have not a demo version, in far 1999 had bought the full version, the demo never used

michaelzangl:
I have both versions here and mixed them up. You are right, the gold edition uses those images. I will have a look at it to find the differences.

PR #84, opened 2015-05-19
Building construction alignment correction

adamjryan:
Removing an offset added to the map draw procedure seemed to correct the
issue. Also removed the previous code that was added to correct the
construction image as it is therefore no longer needed.
There is still a slight issue for some zoom levels when the building is at
a certain height on the screen where the points must be misaligned with
the pixels between two draws, there wasn't an obvious fix for this and
it is not frequently seen, it needs a more in depth investigation.


michaelzangl:
Image displacement be fixed in the current version.

PR #86, opened 2015-05-23
Replaced the quit short-key with a control-panel menu option

adamjryan:


michaelzangl:
Merged in the new menu branch.

PR #89, opened 2015-05-26
Stock control panels updated visually

adamjryan:
An initial working version thrown together for a starting point.
Contains only the visual functionality that would need to be
tied into the game logic.

### DIFF
--- a/jsettlers.graphics/src/jsettlers/graphics/map/controls/original/panel/content/BarFill.java
+++ b/jsettlers.graphics/src/jsettlers/graphics/map/controls/original/panel/content/BarFill.java
@@ -1,0 +1,97 @@
+/*******************************************************************************
+ * Copyright (c) 2015
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ *******************************************************************************/
+package jsettlers.graphics.map.controls.original.panel.content;
+
+import go.graphics.GLDrawContext;
+import jsettlers.common.images.EImageLinkType;
+import jsettlers.common.images.ImageLink;
+import jsettlers.common.images.OriginalImageLink;
+import jsettlers.common.position.FloatRectangle;
+import jsettlers.graphics.action.Action;
+import jsettlers.graphics.action.ExecutableAction;
+import jsettlers.graphics.utils.UIPanel;
+
+public class BarFill extends UIPanel
+{
+    private static final ImageLink barImageLink = new OriginalImageLink(EImageLinkType.GUI, 3, 330, 0); //or 336
+
+    private final UIPanel fill;
+    private final UIPanel frame;
+    private ExecutableAction listener;
+    private float value;
+
+    public BarFill()
+    {
+        fill = new UIPanel(){
+            @Override
+            public void drawAt(GLDrawContext gl) {
+                gl.color(0f, 0.78f, 0.78f, 1f);
+                FloatRectangle position = getPosition();
+                gl.fillQuad(position.getMinX(), position.getMinY(), position.getMaxX(), position.getMaxY());
+            }
+        };
+
+        frame = new UIPanel();
+        frame.setBackground(barImageLink);
+        addChild(frame, 0f, 0f, 1f, 1f);
+    }
+
+    @Override
+    public Action getAction(final float relativex, float relativey)
+    {
+        return new ExecutableAction() {
+            @Override
+            public void execute() {
+                setBarFill(relativex);
+                if (listener != null) {
+                    listener.execute();
+                }
+            }
+        };
+    }
+
+    public void setAction(ExecutableAction action)
+    {
+        this.listener = action;
+    }
+
+    /**
+     *
+     * @return A value from 0 to 1 indicating the proportion of the bar filled.
+     */
+    public float getValue()
+    {
+       return value;
+    }
+
+    /**
+     *
+     * @param percentage expects a value in the range 0.0 to 1.0. Values outside this range
+     * will be clamped.
+     */
+    public void setBarFill(float percentage)
+    {
+        if (percentage < 0) {
+            percentage = 0;
+        }
+        else if (percentage > 1) {
+            percentage = 1;
+        }
+        value = percentage;
+        removeAll();
+        addChild(fill, 0f, 0f, percentage, 1f);
+        addChild(frame, 0f, 0f, 1f, 1f);
+    }
+}
--- a/jsettlers.graphics/src/jsettlers/graphics/map/controls/original/panel/content/DistributionPanel.java
+++ b/jsettlers.graphics/src/jsettlers/graphics/map/controls/original/panel/content/DistributionPanel.java
@@ -1,0 +1,244 @@
+/*******************************************************************************
+ * Copyright (c) 2015
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ *******************************************************************************/
+package jsettlers.graphics.map.controls.original.panel.content;
+
+import go.graphics.text.EFontSize;
+import jsettlers.common.images.EImageLinkType;
+import jsettlers.common.images.OriginalImageLink;
+import jsettlers.graphics.action.ExecutableAction;
+import jsettlers.graphics.localization.Labels;
+import jsettlers.graphics.map.controls.original.panel.content.Label.HorizontalAlignment;
+import jsettlers.graphics.utils.Button;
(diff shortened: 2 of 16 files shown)


jsettlers.graphics/src/jsettlers/graphics/map/controls/original/panel/content/ToolsPanel.java
```diff
@@ -0,0 +1,200 @@
...
+import jsettlers.common.images.ImageLink;
+import jsettlers.common.images.OriginalImageLink;
+import jsettlers.common.map.IGraphicsGrid;
+import jsettlers.common.position.ShortPoint2D;
+import jsettlers.graphics.action.ExecutableAction;
+import jsettlers.graphics.localization.Labels;
+import jsettlers.graphics.utils.Button;
+import jsettlers.graphics.utils.UIPanel;
+
+public class ToolsPanel implements IContentProvider
```
michaelzangl: plaese use AbstractContentProvider of the new controls-improve-menu branch the next time. I will integrate all menu changes into that branch for the moment (until Android is fixed...)

jsettlers.graphics/src/jsettlers/graphics/map/controls/original/panel/content/DistributionPanel.java
```diff
@@ -0,0 +1,286 @@
...
+
+        panel = new UIPanel();
+        panel.addChild(new Label(Labels.getString("controlpanel_distribution_title"), EFontSize.NORMAL), 0f, titleTop - titleTextHeight, 1f, titleTop);
+
+        Item[] items = Item.values();
+        int itemIdx = 0;
+        float top = marginTop;
+        for (int r = 0; r < ROWS; r++, top -= (rowHeight + marginV)) {
+            final Item item = items[itemIdx];
+            Icon icon = item.getIcon();
```
michaelzangl: This returns a reference to a static object.

jsettlers.graphics/src/jsettlers/graphics/map/controls/original/panel/content/DistributionPanel.java
```diff
@@ -0,0 +1,286 @@
...
+            panel.addChild(
+                    items[itemIdx].getIcon(),
+                    marginH,
+                    top - rowHeight,
+                    marginH + tileWidth,
+                    top
+                    );
+            icon.setActionListener(new ExecutableAction() {
```
michaelzangl: Here you are setting an action listenerl. I know this works in this case, but we should at least add a comment that those static icons are used this way (and it would be good to make the DistributionPanel-Class singleton)

jsettlers.graphics/src/jsettlers/graphics/map/controls/original/panel/content/DistributionPanel.java
```diff
@@ -0,0 +1,286 @@
...
+                    }
+                });
+                row.setPercentage(percentage);
+                addChild(row, 0f, top - rowHeight, 1f, top);
+                top -= rowHeight + rowSpacing;
+            }
+        }
+    }
+
+    private static class Icon extends UIPanel
```
michaelzangl: Nice idea to have this. If you want to, you can merge this with our "Button" class and move it to that package.

michaelzangl:
Great work. I added some comments on the details, but the rest looks fine and well structured.

jsettlers.graphics/src/jsettlers/graphics/map/controls/original/panel/content/DistributionPanel.java
adamjryan: Yeah came across a few variants that could probably be combined into a more configurable Button class.

jsettlers.graphics/src/jsettlers/graphics/map/controls/original/panel/content/DistributionPanel.java
adamjryan: They're private, essentially it was just supposed to be a convienient immutable list, I'll look at something a bit tidier.

jsettlers.graphics/src/jsettlers/graphics/map/controls/original/panel/content/ToolsPanel.java
adamjryan: It'll probably be best it this goes in after that branch is merged, and I'll update this to use the new stuff.

adamjryan:
Cheers Ivan, I spotted them through MaterialPriorityPanel -> EMaterialType but forgot to add them. Will probably need to update to use that resource instead of the repeating the definitions.

jsettlers.graphics/src/jsettlers/graphics/map/controls/original/panel/content/DistributionPanel.java
michaelzangl: The problem is that they are not immutable as soon as you start setting an event listener every time. You never use the enum names, so you could simply use a normal List of Objects (it does not have to be static, you can dynamically create it every time the panel is opened).

justangel:
Hi @adamjryan
great job, but I think names needed in the general style for easy search and less confusion

jsettlers.graphics/src/jsettlers/graphics/map/controls/original/panel/content/BarFill.java
```diff
@@ -0,0 +1,89 @@
...
+                    listener.execute();
+                }
+            }
+        };
+    }
+
+    @Override
+    public Action getAction(float relativex, float relativey)
+    {
+        value = relativex;
```
michaelzangl: This does not need to be caused by a mouse event. You should not set value here.

jsettlers.graphics/src/jsettlers/graphics/map/controls/original/panel/content/BarFill.java
```diff
@@ -0,0 +1,89 @@
...
+     * @return A value from 0 to 1 indicating the proportion of the bar filled.
+     */
+    public float getValue()
+    {
+       return value;
+    }
+
+    public void setBarFill(float percentage)
+    {
+        removeAll();
```
michaelzangl: Just add a value = percentage

jsettlers.graphics/src/jsettlers/graphics/map/controls/original/panel/content/DistributionPanel.java
```diff
@@ -0,0 +1,286 @@
...
+import jsettlers.graphics.utils.UIPanel;
+
+public class DistributionPanel implements IContentProvider {
+    private static class ConfigurationPanelRow extends UIPanel
+    {
+        private static final float textHeight = rowTextHeight_px / rowHeight_px;
+        private static final float textMarginBottom = rowTextMarginBottom_px / rowHeight_px;
+        private static final float textPercentageWidth = 30f / 84f;
+
+        Label lblPercentage;
```
michaelzangl: You don't need to add those Objects as private (!) fields if they are only used once.

jsettlers.graphics/src/jsettlers/graphics/map/controls/original/panel/content/DistributionPanel.java
```diff
@@ -0,0 +1,286 @@
...
+        BarFill barFill;
```
michaelzangl: private ...

jsettlers.graphics/src/jsettlers/graphics/map/controls/original/panel/content/DistributionPanel.java
```diff
@@ -0,0 +1,286 @@
...
+        }
+
+        @Override
+        public Action getAction(float relativex, float relativey)
+        {
+            return actionListener;
+        }
+    }
+
+    private enum Item {
```
michaelzangl: I know this is mostly for testing. We definitely need a way to get this from the game logic.

jsettlers.graphics/src/jsettlers/graphics/map/controls/original/panel/content/InventoryPanel.java
```diff
@@ -0,0 +1,161 @@
...
+            goodsIcon.setBackground(imageLink);
+            addChild(goodsIcon, 0f, quantityTextTop, 1f, 1f);
+
+            lblQuantity = new Label(Labels.getString("0"), EFontSize.SMALL);
+            addChild(lblQuantity, 0f, 0f, 1f, quantityTextTop);
+        }
+    }
+
+    private enum Item {
+        PLANK(new InventoryItem(new OriginalImageLink(EImageLinkType.GUI, 3, 168, 0))),
```
michaelzangl: All this information is already present in EMaterialType.

jsettlers.graphics/src/jsettlers/graphics/map/controls/original/panel/content/Label.java
```diff
@@ -33,6 +39,16 @@ public Label(String message, EFontSize size) {
 		words = message.split(" ");
```
michaelzangl: replace this with setText()

jsettlers.graphics/src/jsettlers/graphics/map/controls/original/panel/content/Label.java
```diff
@@ -20,12 +20,18 @@
...
 public class Label extends UIPanel {
+    public enum HorizontalAlignment {
+        LEFT,
+        CENTRE,
+        RIGHT,
+    }
 
 	private final EFontSize size;
-	private final String[] words;
+	private String[] words;
```
michaelzangl: If words is not final, you need to synchronize access to it.

michaelzangl:
Great work.

I would definitely want to merge some stuff of this. At least the button descriptions (so that they get cleickable) and your changes to UIPanel/Label. If you want to, you can create a new pull request with those changes and I will merge them now.

For the rest, I would like to keep this open and wait for the logic to be implemented. We can then add them step by step.

michaelzangl:
If you want to implement some features, you can have a look at the android code and the IPartitionSettings interface.

adamjryan:
It's great to see things progressing and thanks for the feedback, I'll look to sort this one out in the next few days.

adamjryan:
I've made the changes and the actual additions to the front end are now commented out (not visible) so this can be merged and uncommented as the back end is wired in.

jsettlers.graphics/src/jsettlers/graphics/map/controls/original/panel/content/BarFill.java
adamjryan: It's supposed to be a response to the mouse event, I'm not sure what you mean.

jsettlers.graphics/src/jsettlers/graphics/map/controls/original/panel/content/Label.java
adamjryan: Front ends are typically single threaded to keep it responsive and avoid lock ups. I think adding synchronisation blocks might encourage the wrong type of architecture.

jsettlers.graphics/src/jsettlers/graphics/map/controls/original/panel/content/BarFill.java
michaelzangl: getAction() may not be the result of a real mouse click. Actions can be canceled/replaced/... and this method might even be used in the future to select the right cursor to use and thus not only called when a mouse button is pressed

jsettlers.graphics/src/jsettlers/graphics/map/controls/original/panel/content/Label.java
michaelzangl: You always need a point where you synchronize with the backend, since that is not aware of any UI thread scheduling. Synchronisation does not have to be here, but this is the point were it crashes.

jsettlers.graphics/src/jsettlers/graphics/map/controls/original/panel/content/BarFill.java
adamjryan: Ah, makes sense.

jsettlers.graphics/src/jsettlers/graphics/map/controls/original/panel/content/Label.java
adamjryan: I suppose I'm suggesting that the UI should have an event queue for events from the back end, which is probably a bit of a change for now, I've put something a bit simpler in.

michaelzangl: We can't do this for this game. There are so many changes to the game grid by logic that we cannot afford to pass them all through a queue system. And we cannot afford to pause logic every time the graphic gets drawn. This is why we accept that the UI is not completely synchronized (as it would be in a well-written Swing application) but instead we accept that for a small fraction of a second, an "invalid" state in the middle of an update might be shown. We often have the case that movables at the top of the screen are one tick behind the ones at the bottom. The user won't see the difference. But we need to ensure that no crashes (Array out of bounds, ...) result from this.

adamjryan: In case you missed it, I did add the change to avoid any concurrency issues in the last update so it should be good to go.

andreas-eberle:
Instead of merging this into master with commented out lines, I'd prefer to continue work on this in a branch in the main repository. This way we can finish it / wire it with the game logic and merge it as a complete feature.

I will open that branch the next days.

@adamjryan: Many thanks! The menus look great!

adamjryan:
Guess I'll have to finish the job then. :-)

andreas-eberle:
I close this PR in favor of the new #112.

PR #90, opened 2015-05-27
Replaced the default cursor graphic with the Settlers themed version.

adamjryan:


andreas-eberle:
I think we should not render the Cursor ourselves, but let Java do this task for us. However, to be able to do this, we need the cursor image as java.awt.Image. We already had code doing this in the DatFileTester. Therefore I moved that code to a method in SingleImage. See my commit ef2f29cbe75dbdb7fd52d53f569c5fa1f6d47aee. 

With this change it would be enough to append the following code to the end of `SwingManagedJSettlers.startJogl()`.

```
		Image cursor = ImageProvider.getInstance().getImage(new OriginalImageLink(EImageLinkType.SETTLER, 2, 20, 0));
		BufferedImage cursorImage = ((SingleImage) cursor).generateBufferedImage();
		Cursor settlersCursor = Toolkit.getDefaultToolkit().createCustomCursor(cursorImage, new Point(0, 0), "Settlers Cursor");
		jsettlersWnd.setCursor(settlersCursor);
```

However, as I explained in #90, it would be great to change the cursor depending on what the user is doing (for example scrolling in the map). I don't exactly know where's the best place to put this code for that. That's why I opened the ticket to discuss the topic first. 

Therefore, I think this PR is not valid any more. Nevertheless, many thanks for the code to set the cursor image.

adamjryan:
There's definitely pros and cons. Using swing is probably better for responsiveness, using JOGL would be better for platform independence and it would be easier to tie it into the platform independent panels etc. Maybe there is some native handling for cursors in JOGL?

PR #91, opened 2015-05-30
Support reading f8007e01f.dat files

justangel:
I spent the time to research and implemented support for reading files "f8007e01f"
transmitting the data in the format of 5-5-5-1

TODO:
if not available 7c003e01f file, search and use f8007e01f or vice versa


jsettlers.graphics/src/jsettlers/graphics/reader/translator/SettlerTranslator.java
```diff
@@ -44,9 +44,22 @@ public HeaderType getHeaderType() {
 	}
 
 	@Override
-	public short readUntransparentColor(ByteReader reader)
-			throws IOException {
-		return (short) ((reader.read16() << 1) | 0x01);
+	public short readUntransparentColor(ByteReader reader) throws IOException {
+		int read = reader.read16();
+
+		int R5 = (int)((read & 0xF800) >> 11);
```
michaelzangl: Can you please move this to a new method (e.g. convert565to555).

michaelzangl:
Nice work. Are you sure that the f8007e01f-Files are shipped with all versions of settlers? If yes, we could apply your patch. Otherwise, we should use an if/else for this.

I would also like to check if there are performance changes when reading the file (and change your float ops to int ops if they are a problem).

justangel:
I have a gold edition, where both versions of these files
I think it is better to use check if/else
It will not be worse, but the chance to work will be higher

jsettlers.graphics/src/jsettlers/graphics/reader/translator/SettlerTranslator.java
justangel: https://github.com/justangel/settlers-remake/commit/aabfc02ce8cf5684a3e97eb05ac2bbb49f3fce4c right? or other place better to move?

michaelzangl: Looks fine, although I prefer lower case variable names for non-constants ;-).

I don't think that the ByteReader is the best place to put that. We could add it as a read-and-convert method. But the way it is now I would rather see it as a static method in one of the image-reading related classes, since the byte reader has nothing to do with the interpretation of images. You can also create a new class (ColorUtils, ...) or simply add it as static method to our Color-Class (which also does 16 Bit to 24 Bit conversion, ...)

justangel: moved to Color as static https://github.com/jsettlers/settlers-remake/commit/6f07449288aa3ba3ec54fe2fbef31f710d17980d

justangel:
in my GFX folder are located 45 files of 7c003e01f format and 47 files of f8007e01f
needed to use if/else for selecting and reading files

I will try to implement it, I recently started to learn java language, just with the release of this project as an open source =)

justangel:
in Gold edition v1.60 i deleted all 7c003e01f files and run the game, all works fine
when i deleted all f8007e01f files and try to run the game - prompted to insert the CD

justangel:
the results:
no files 7c003e01f - game works
no files f8007e01f - game does not work

It means f8007e01f basic files

michaelzangl:
OK, then we would not require a dnamic switch. I would like to keep the old code working (just to be sure) but we can just add a constant in our constants class (in the commons project).

You can then just do an if(CommonConstants.USE_ALTERNATIVE_IMAGE_FORMAT) { old code} else {new code} ;-)

The nicer way would be to implement them both in an object-oriented way, but I think this is overkill here.

michaelzangl:
Merged in 852168566042f6e65f61b351220d36acd5f55ee7

justangel:
closed by mistake?

michaelzangl:
No, just wrong commit number. The right one is: b33508b90d7180834f874496682e16ac24d3834d

justangel:
ok, thanks, great work

PR #93, opened 2015-06-05
Retina Support for OSX

codingberlin:
A Mac could have a retina display device where the GUI is very sharp or scaled up.
This behavior could lead into wrong mouse position and click calculation so that the user can't interact with the GUI.

First issue:
Apple JRE supports scaling until including version 1.6.x
Oracle JRE does not support scaling until 1.7.0_40

Second issue:
You can connect a non-retina device to your retina Mac and move the settlers window around. At the retina device you need to scale the mouse position at the non-retina device you must not.

Solution:
You need to ask the running java.awt.Window at which device it runs and figure out the scaling factor of this device.

I testet the code with JRE 1.6.0_65, 1.7.0_25, 1.7.0_79, 1.8.0_45 on my Mac connected to a non-retina TV and moved around the Window.


go.graphics.swing/src/go/graphics/swing/event/swingInterpreter/GOSwingEventConverter.java
```diff
@@ -45,6 +49,8 @@
 	 */
 	private boolean panWithButton3;
 
+	private static Window referenceWindowToCalculateRetinaScale;
+
```
andreas-eberle: It's always a good idea to try not to create static fields, as this creates a state in the game, which cannot easily be reset. Instead of setting the Window to this field, you could get the GraphicsConfiguration from the Component given to the constructor.

Therefore I think it's best to calculate the scale factor only once in the constructor and save the factor in a (non-static) field. This also ensures that no other class has to know about this implementation detail of this class. Especially SwingManagedJSettlers does not have to be modified / even know about this class.

Can you do the changes? Besides that: Thanks for your contribution!

codingberlin:
Thanks for the hint. I didn't noticed the Component parameter in the constructor.

Furthermore I reduced the times of calculation of the retinaScaleFactor of the start and movement of the window.

It was necessary to add ComponentListener and HierarchyListener in order to get the parent Window because it is not available when the constructor is called yet.

michaelzangl:
Is this stable and ready to be merged?

codingberlin:
In my opinion it is stable and ready to be merged.

I solved the issues from andreas-eberle's comment.

I tested the code with JRE 1.6.0_65, 1.7.0_25, 1.7.0_79, 1.8.0_45 on my Mac connected to a non-retina TV and moved around the Window.

PR #112, opened 2015-06-17
Feature: Implement stock control / information pannels

andreas-eberle:
This is the new branch for continuing work started in PR #89.


jsettlers.common/src/jsettlers/common/map/partition/IPartitionData.java
```diff
@@ -41,4 +41,13 @@
...
 	EMaterialType getMaterialTypeForPrio(int priorityIdx);
+
+	/**
+	 * Returns the number of materials of the given type, available in the respective partition.
+	 * 
+	 * @param materialType
+	 *            Type of material.
+	 * @return Available amount of that material.
+	 */
+	int getAmountOf(EMaterialType materialType);
```
andreas-eberle: @adamjryan: You can use this method to get the number of materials available in the current partition. To get an instance of this interface, have a look at `IGraphicsGrid.getPartitionData(int x, int y):IPartitionData`.

At the moment `getAmountOf()` always returns 0. But I will implement it soon. In the meantime, you can start wiring it into the UI.

Furthermore, the method `IPartitonData.getDistributionSettings(EMaterialType materialType)` provides the material distribution for the buildings (i.e. what percentage of a material is going to each type of consumer building). This method should provide correct values already.

michaelzangl:
@adamjryan: If you want to, you can have a look at the jsettlers.graphics/res/layout directory. There is an example layout file. The layout system is still work in progress, but you can have a look at the building selection dialog. I want to use it for all layouts that have many of positioned children.

adamjryan:
Yeah looks good, it will be interesting to try it out, didn't expect to see it come together so easily! I'm not going to have chance at the weekend so I'll get back into it next week.

michaelzangl:
@andreas-eberle Can you have a look at the autoplay changes?

andreas-eberle:
To prevent further merge conflicts with the master, I merged this branch already. Refinements of the panels and further connection between UI and game logic will be done in new branches,

Issue #117, opened 2015-06-18
Show global and partiton building count in construction tooltip


michaelzangl:
The tooltip should look like:

> Lumberjack
> In this partition: 5
> In this world: 7 (+1)

andreas-eberle:
The logic has been implemented in #124. The UI can be implemented now.

andreas-eberle:
The building count has already been added to the menu. Can you also add the tooltip?

PR #124, opened 2015-06-22
Display number of buildings in construction menu (implements #117)

andreas-eberle:


andreas-eberle:
The game logic now calculates the number of buildings of a player and in the current partition. The interface IPartitionData has been adapted by adding a new method providing this data.

@michaelzangl: Can you do the UI implementation?

PR #139, opened 2015-08-30
Computerplayer

codingberlin:
Siedler3 Remake Computerplayer

# Features:
- the following buildings will be build:
    - LUMBERJACK
    - SAWMILL
    - FORESTER
    - STONECUTTER
    - WINEGROWER
    - FARM
    - TEMPLE
    - WATERWORKS
    - MILL
    - BAKER
    - PIG_FARM
    - SLAUGHTERHOUSE
    - COALMINE
    - IRONMINE
    - IRONMELT
    - WEAPONSMITH
    - BARRACK
    - STOCK
    - BIG_TEMPLE
    - FISHER
    - GOLDMINE
    - GOLDMELT
    - WATERWORKS
    - PIG_FARM 
    - TOWER
    - SMALL_LIVINGHOUSE
    - MEDIUM_LIVINGHOUSE
- when a tower is finished a new will be build
- towers will be build in direction: trees, stones, rivers, coal, iron, fish, gold (in that priority)
- towers surrounded by other towers will be destroyed to get back soldiers and materials
- a solder is sent to the tower construction side
- if no soldier was available when a new tower was placed or the solder which walked to the tower was killed, a new soldier will walk to the tower
- tool smith with its dependencies will be build as soon as any tool is missing for the next building
- living houses will be build before there are too less settlers
- stonecutters will be destroyed when the stone is empty
- lumberjacks, farms and winegrowers get space in the south where no other building will be placed so that they not run out of resources
- fishers will be placed near by fish
- lumberjacks will be placed near by trees
- foresters will be placed near by lumberjacks but not near by other foresters
- mines will be placed on their resources
- living houses will be placed near by diggers
- all other buildings will be placed near their dependent buildings (e.g. bakery near by a farm)
- storages will be build when the gold amount increases
- the computer player is connected to the game clock which is necessary to use F12 and the multiplayer
- the actions of the computer player are wrapped in tasks which is necessary for the multiplayer

# Still open:
- effort to flatten land must be calculated negative to the score when a place for a new building is searched
- winegrowers need to be placed in the north of hills
- the enemy needs to be attacked by the computer player when enough soldiers are present
- crush living houses when enough settlers were transformed to soldiers to get back materials
- a defense weapon smith needs to be build when the enemy is a rusher
- fight AI needs to be implemented like with the strategy of: http://tips.siedler3.net/page.php?lan=de&article=kampfkunst (german)
- looping work area change to increase efficiency (e.g. the forrester can plant new trees in the green area when the lumberjack cleared it, instead of planting trees in the red area)
- use market and donkeys instead of storages for gold, because this saves material
- better algorithm to place mines in order to build maximum number of mines for the late game and not only a few 100% mines

# Technical details
- there is a high level KI to decide WHAT to do
- there is a low level KI to decide WHERE to do
- there is a AiStatistics class which collects data from the grids and is the interface for the high and low-level ki
- currently every player which is no human player will be one special computer player: RomanWhatToDoAi -> later it could be possible to create different AIs which can be chosen before game start. e.g. easy, medium and hard player, rusher, economist
- NetworkTimer can now notify listeners. This was chosen instead of sending AiExecutionTasks around the network because this was heavy IO and reduced the game performance
- Ai is executed only at the host computer
- I oriented at the YAGNI principle and implemented O(n²) first and take some measurements with jvisualvm and after that I implemented O(n*log n) only for the hot spots. The performance is ok at my mac. I don’t have experience with android. please give me feedback if there is an issue, than I keep optimizing (which will increase the complexibility of the code)

# What I noticed:
- running the AI helps to detect issues in the balance:
- insufficient number of solders or goods for map creation
- maybe to long time to produce goods (e.g corn)
- toolsmith: hammer and shovels should not be produced when other tools are needed. (e.g. the scythe is more imported because it leads to food which leads to iron ore and coal which leads to much more tools)
- no goods (especially gold) are not stored in stocks

Feedback is strong welcome!


codingberlin:
I noticed that "runAutoReplayIT" did not pass. I'll check this another day because today my coding time is up ;)

michaelzangl:
This means that you changed the game in a way that influences the random state.

Fell free to read into that test case. it is mainly used to detect influences of UI changes or code refactoring to the game logic.

jsettlers.common/src/jsettlers/common/position/ShortPoint2D.java
```diff
@@ -114,4 +114,11 @@ public static int getOnGridDist(final int dx, final int dy) {
 	public boolean equals(int x, int y) {
 		return this.x == x && this.y == y;
 	}
+
+	public double calculateDistanceTo(ShortPoint2D that) {
+		if (that == null) {
+			return 0;
+		}
+		return Math.sqrt((this.x - that.x) * (this.x - that.x) + (this.y - that.y) * (this.y - that.y));
```
michaelzangl: This is not the distance of two points, since we are working on a screwed grid.

Have a look at
jsettlers.common.map.shapes.MapCircle#getSquaredDistance(dx, dy)

jsettlers.logic/src/jsettlers/ai/construction/BestMilitaryConstructionPositionFinder.java
```diff
@@ -0,0 +1,178 @@
...
+					scoredConstructionPositions.add(new ScoredConstructionPosition(point, score));
+				}
+			}
+		}
+
+		return ScoredConstructionPosition.detectPositionWithLowestScore(scoredConstructionPositions);
+	}
+
+	private List<ImportantResource> detectMostImportantResourcePoints(AiStatistics aiStatistics, byte playerId, ShortPoint2D referencePoint) {
+		List<ImportantResource> importantResources = new ArrayList<ImportantResource>();
```
michaelzangl: You can use an EnumSet if you do not need this to be ordered.

jsettlers.logic/src/jsettlers/ai/highlevel/AiStatistics.java
```diff
@@ -0,0 +1,612 @@
...
+ * multiple times within one AiExecutor step triggerd by the game clock.
+ * 
+ * @author codingberlin
+ */
+public class AiStatistics {
+
+	private static final short BORDER_LAND_WIDTH = 10;
+
+	private final Queue<Building> buildings;
+	private Map<Integer, Map<EBuildingType, Integer>> totalBuildingsNumbers;
```
michaelzangl: You have lots of those playerId => map maps. Consider changing the order of those mappings to use a playerId => AiPlayerStatistics map. This new class can then hold the totalBuildingsNumbers,  buildingsNumbers, ... - Arrays.

jsettlers.logic/src/jsettlers/ai/highlevel/AiStatistics.java
```diff
@@ -0,0 +1,612 @@
...
+	private Map<Integer, Map<EBuildingType, Integer>> unoccupiedBuildingsNumbers;
+	private Map<Integer, Map<EMaterialType, Integer>> materialNumbers;
+	private Map<Integer, Map<EMovableType, List<ShortPoint2D>>> movablePositions;
+	private Map<Integer, Integer> numberOfNotFinishedBuildings;
+	private Map<Integer, Integer> numberOfTotalBuildings;
+	private Map<Integer, Integer> numberOfNotOccupiedTowers;
+	private Map<Integer, List<ShortPoint2D>> stones;
+	private Map<Integer, List<ShortPoint2D>> trees;
+	private Map<Integer, List<ShortPoint2D>> rivers;
+	private Map<EMapObjectType, Map<Integer, List<Integer>>> sortedCuttableObjectsInDefaultPartition;
```
michaelzangl: Iteration over this map is not really sorted.

Mind that those maps might get big really fast. Often, you are faster and less memory-consuming by using bitsets. You can change this later if you are only iterating over your lists and never rely on them of being real ShortPoint2D lists.

jsettlers.logic/src/jsettlers/ai/highlevel/AiStatistics.java
```diff
@@ -0,0 +1,612 @@
...
+			return getNearestPoinInDefaultPartitionOutOfSortedMapInYDirection(ypsilons, point, currentNearestPointDistance, x, y + increment,
+					increment, border, playerId);
+		}
+		if (point.calculateDistanceTo(new ShortPoint2D(x, y)) > currentNearestPointDistance) {
+			return null;
+		}
+		return new ShortPoint2D(x, y);
+	}
+
+	public List<ShortPoint2D> getMovablePositionsByTypeForPlayer(EMovableType movableType, byte playerId) {
```
michaelzangl: I suggest you use IMapArea as return type here. You can write a simple wrapper class for your list.

jsettlers.logic/src/jsettlers/ai/highlevel/AiStatistics.java
```diff
@@ -0,0 +1,612 @@
...
+		Integer playerIdInteger = new Integer(playerId);
```
michaelzangl: Please never use new Integer(). Always use Integer.valueOf(playerId). It is faster.

jsettlers.logic/src/jsettlers/ai/highlevel/AiStatistics.java
michaelzangl: You know the maximum number of players on the map. An array with that length is more efficient than a hash table and does not consume more memory (since the player count is low)

jsettlers.logic/src/jsettlers/ai/highlevel/AiStatistics.java
```diff
@@ -0,0 +1,612 @@
...
+					if (!trees.containsKey(playerId)) {
+						trees.put(playerId, new ArrayList<ShortPoint2D>());
+					}
+					if (objectsGrid.hasCuttableObject(x, y, EMapObjectType.TREE_ADULT)) {
+						trees.get(playerId).add(point);
+					}
+					if (!rivers.containsKey(playerId)) {
+						rivers.put(playerId, new ArrayList<ShortPoint2D>());
+					}
+					if (landscapeGrid.getLandscapeTypeAt(x, y) == RIVER1 || landscapeGrid.getLandscapeTypeAt(x, y) == RIVER2
```
michaelzangl: Consider using:
RIVERS = EnumSet.of(RIVER1, RIVER2, ...);
RIVERS.contains(landscapeGrid.getLandscapeTypeAt(x, y))

jsettlers.logic/src/jsettlers/logic/map/grid/MainGrid.java
```diff
@@ -853,12 +885,16 @@ public final short getHeight() {
...
 
-		boolean canConstructAt(short x, short y, EBuildingType type, byte playerId) {
+		@Override
+		public boolean canConstructAt(short x, short y, EBuildingType type, byte playerId) {
 			ELandscapeType[] landscapes = type.getGroundtypes();
 			RelativePoint[] protectedTiles = type.getProtectedTiles();
 
 			int firstPosX = protectedTiles[0].calculateX(x);
 			int firstPosY = protectedTiles[0].calculateY(y);
+			if (firstPosX < 0 || firstPosY < 0 || firstPosX > partitionsGrid.getWidth() || firstPosY > partitionsGrid.getHeight()) {
```
michaelzangl: We have an inMap()-Method for this somewhere ;-).

And we should check this for the whole building, not just the start position.

jsettlers.logic/src/jsettlers/main/JSettlersGame.java
```diff
@@ -93,6 +97,7 @@ private JSettlersGame(IGameCreator mapCreator, long randomSeed, INetworkConnecto
 		this.networkConnector = networkConnector;
 		this.playerId = playerId;
 		this.availablePlayers = availablePlayers;
```
michaelzangl: Idea: Change this to a PlayerSetting[] availablePlayers. Then give them an isAvailable() and an isAi() method. In the future, we could add difficulty/... flags here.

jsettlers.network/src/jsettlers/network/synchronic/timer/NetworkTimer.java
```diff
@@ -324,7 +331,7 @@ public void setTime(int newTime) {
 	}
 
 	@Override
-	public int getTime() {
+	public synchronized int getTime() {
```
michaelzangl: This should never be used outside the main game loop.

jsettlers.logic/src/jsettlers/ai/highlevel/AiStatistics.java
```diff
@@ -0,0 +1,612 @@
...
+
+		ShortPoint2D nearestSoldierPosition = detectNearestPointFromList(targetPosition, soldierPositions);
+		return movableGrid.getMovableAt(nearestSoldierPosition.x, nearestSoldierPosition.y);
+	}
+
+	public ShortPoint2D detectNearestPointFromList(ShortPoint2D referencePoint, List<ShortPoint2D> points) {
+		ShortPoint2D nearestPoint = null;
+		double nearestPointDistance = Double.MAX_VALUE;
+		for (ShortPoint2D point : points) {
+			double currentPointDistance = referencePoint.calculateDistanceTo(point);
```
michaelzangl: Use the calculateDistanceSquared(...)-Methods. They are faster and the result is the same.

jsettlers.logic/src/jsettlers/ai/highlevel/AiStatistics.java
```diff
@@ -0,0 +1,612 @@
...
+			if (nearestPoint == null || currentPointDistance < nearestPointDistance) {
```
michaelzangl: currentPointDistance < nearestPointDistance is always the case if nearestPointDistance == Double.MAX_VALUE (I always use Double.POSITIVE_INFINITY to make debugging easier)

jsettlers.logic/src/jsettlers/ai/highlevel/AiStatistics.java
```diff
@@ -0,0 +1,612 @@
...
+ */
+public class AiStatistics {
+
+	private static final short BORDER_LAND_WIDTH = 10;
+
+	private final Queue<Building> buildings;
+	private Map<Integer, Map<EBuildingType, Integer>> totalBuildingsNumbers;
+	private Map<Integer, Map<EBuildingType, Integer>> buildingsNumbers;
+	private Map<Integer, Map<EBuildingType, Integer>> unoccupiedBuildingsNumbers;
+	private Map<Integer, Map<EMaterialType, Integer>> materialNumbers;
```
michaelzangl: You can use an int[EMaterialType.values().length()] here. A hash table entry takes 8 Byte (entry pointer in the Hashtable) + 8 Byte (Object header) + 8 Byte (EMaterialType-Pointer) + 8 Byte (Integer-Pointer) + some more bytes for internal use. So an int[] is even more memeory-efficient in most normal cases.

jsettlers.logic/src/jsettlers/ai/highlevel/AiStatistics.java
```diff
@@ -0,0 +1,612 @@
...
+		return flattenEffort;
+	}
+
+	public void updateStatistics() {
+		updateBuildingStatistics();
+		updateMapStatistics();
+	}
+
+	private void updateBuildingStatistics() {
+		totalBuildingsNumbers = new HashMap<Integer, Map<EBuildingType, Integer>>();
```
michaelzangl: You can simply use totalBuildingsNumbers.clear()

jsettlers.logic/src/jsettlers/ai/highlevel/RomanWhatToDoAi.java
```diff
@@ -0,0 +1,453 @@
...
+					sendMovableTo(soldier, door);
+				}
+			}
+		}
+	}
+
+	private void sendMovableTo(Movable movable, ShortPoint2D target) {
+		List<Integer> selectedIds = new ArrayList<Integer>();
+		if (movable != null) {
+			selectedIds.add(movable.getID());
```
michaelzangl: Use Collections#singletonList(movable.getID())

michaelzangl:
As you can see, I went through the source code and added some notes. All in all, your work is amazing so far. Most of the comments are simple speed or readability improvements. The more you write, the harder it gets to change that later. But don't sacrifice readability for it ;-).

Feel free to ask if you need any help with coding, algorithms (like fast scanning, ...) or with our source code. We already have methods that compute the distance taking into account the hex grid (both the screen distance and the walking distance) or to iterate over the map or parts of it.

codingberlin:
All tests passes now.
The problem was a deadlock caused by ```synchronized```

I merged the current master.

Next I will go through the comments and solve them.

jsettlers.logic/src/jsettlers/ai/highlevel/AiStatistics.java
codingberlin: done

jsettlers.logic/src/jsettlers/ai/highlevel/AiStatistics.java
codingberlin: Done.
Thx for the hint. I just didn't notice the existing better distance methods.

jsettlers.logic/src/jsettlers/logic/map/grid/MainGrid.java
andreas-eberle: It is MainGrid.isInBounds(x,y)

andreas-eberle:
As discussed with @codingberlin, I pushed his branch to our repository (feature_computerplayer). This PR will stay open until the comments are fixed. After that, we will create a new PR for the new branch in our repository. 

No further commits should therefore be made to this branch.

jsettlers.logic/src/jsettlers/logic/map/grid/MainGrid.java
codingberlin: I changed the code as following:
```if (!isInBounds(firstPosX, firstPosY)) {
				return false;
			}
			final short partitionId = getPartitionIdAt(firstPosX, firstPosY);
```
I only check the first x,y pair because in the next line it would lead to an ArrayOutOfBounds Exception when the partitionID is requested.

Some lines below all relative positions were checked against inBounds by calling canUsePositionForConstruction

```
for (RelativePoint curr : protectedTiles) {
				int currX = curr.calculateX(x);
				int currY = curr.calculateY(y);

				if (!canUsePositionForConstruction(currX, currY, landscapes, partitionId)) {
					return false;
				}
			}
```

```
public boolean canUsePositionForConstruction(int x, int y, ELandscapeType[] landscapeTypes, short partitionId) {
			return isInBounds(x, y)
					&& !flagsGrid.isProtected(x, y)
					&& partitionsGrid.getPartitionIdAt(x, y) == partitionId
					&& isAllowedLandscape(x, y, landscapeTypes);
		}
```

codingberlin:
As discussed with @andreas-eberle, I removed the IClockListener because Timerable does the same.

jsettlers.common/src/jsettlers/common/position/ShortPoint2D.java
codingberlin: I removed the message because I used the established methods now.

jsettlers.logic/src/jsettlers/ai/construction/BestMilitaryConstructionPositionFinder.java
codingberlin: Done

jsettlers.logic/src/jsettlers/main/JSettlersGame.java
codingberlin: I introduced the PlayerSetting[] playerSettings now.
Please take a look at the code: Is it like meant it now?

jsettlers.logic/src/jsettlers/ai/highlevel/AiStatistics.java
codingberlin: I introduced PlayerStatistic class. It's instances are stored in an array.
It is not called AiPlayerStatistic because it holds statistics from human players as well.
This is needed later when the ai decides to attack or not to attack by the number of soldiers of the human player.

jsettlers.logic/src/jsettlers/ai/highlevel/AiStatistics.java
codingberlin: ```currentPointDistance < nearestPointDistance```is only true for the first iteration. 2 lines below nearestPointDistance is modified.

```
int nearestPointDistance = Integer.MAX_VALUE;
		for (ShortPoint2D point : points) {
			int currentPointDistance = referencePoint.getOnGridDistTo(point);
			if (nearestPoint == null || currentPointDistance < nearestPointDistance) {
				nearestPoint = point;
				nearestPointDistance = currentPointDistance;
			}
		}
```

jsettlers.logic/src/jsettlers/ai/highlevel/AiStatistics.java
codingberlin: Done

jsettlers.logic/src/jsettlers/ai/highlevel/AiStatistics.java
codingberlin: Done

jsettlers.logic/src/jsettlers/ai/highlevel/RomanWhatToDoAi.java
codingberlin: Done

jsettlers.logic/src/jsettlers/ai/highlevel/AiStatistics.java
codingberlin: I don't understand your suggestion. I think an MapArea should be connected but the returned positions of the method will be dappled and not connected.

jsettlers.logic/src/jsettlers/ai/highlevel/AiStatistics.java
codingberlin: this sorted* lists are used to do a "nearest neighbor search" algorithm which does not visit every member in a list to reach a lower complexity as O(n*n)

jsettlers.logic/src/jsettlers/ai/highlevel/AiStatistics.java
codingberlin: Done

codingberlin:
I went through all the comments and as discussed close this pull request and create a new one from the new branch in this repository.

PR #146, opened 2015-09-06
Feature computerplayer

codingberlin:
# Features:
- the following buildings will be build:
    - LUMBERJACK
    - SAWMILL
    - FORESTER
    - STONECUTTER
    - WINEGROWER
    - FARM
    - TEMPLE
    - WATERWORKS
    - MILL
    - BAKER
    - PIG_FARM
    - SLAUGHTERHOUSE
    - COALMINE
    - IRONMINE
    - IRONMELT
    - WEAPONSMITH
    - BARRACK
    - STOCK
    - BIG_TEMPLE
    - FISHER
    - GOLDMINE
    - GOLDMELT
    - WATERWORKS
    - PIG_FARM 
    - TOWER
    - SMALL_LIVINGHOUSE
    - MEDIUM_LIVINGHOUSE
- when a tower is finished a new will be build
- towers will be build in direction: trees, stones, rivers, coal, iron, fish, gold (in that priority)
- towers surrounded by other towers will be destroyed to get back soldiers and materials
- a solder is sent to the tower construction side
- if no soldier was available when a new tower was placed or the solder which walked to the tower was killed, a new soldier will walk to the tower
- tool smith with its dependencies will be build as soon as any tool is missing for the next building
- living houses will be build before there are too less settlers
- stonecutters will be destroyed when the stone is empty
- lumberjacks, farms and winegrowers get space in the south where no other building will be placed so that they not run out of resources
- fishers will be placed near by fish
- lumberjacks will be placed near by trees
- foresters will be placed near by lumberjacks but not near by other foresters
- mines will be placed on their resources
- living houses will be placed near by diggers
- all other buildings will be placed near their dependent buildings (e.g. bakery near by a farm)
- storages will be build when the gold amount increases
- the computer player is connected to the game clock which is necessary to use F12 and the multiplayer
- the actions of the computer player are wrapped in tasks which is necessary for the multiplayer

# Still open:
- effort to flatten land must be calculated negative to the score when a place for a new building is searched
- winegrowers need to be placed in the north of hills
- the enemy needs to be attacked by the computer player when enough soldiers are present
- crush living houses when enough settlers were transformed to soldiers to get back materials
- a defense weapon smith needs to be build when the enemy is a rusher
- fight AI needs to be implemented like with the strategy of: http://tips.siedler3.net/page.php?lan=de&article=kampfkunst (german)
- looping work area change to increase efficiency (e.g. the forrester can plant new trees in the green area when the lumberjack cleared it, instead of planting trees in the red area)
- use market and donkeys instead of storages for gold, because this saves material
- better algorithm to place mines in order to build maximum number of mines for the late game and not only a few 100% mines

# Technical details
- there is a high level KI to decide WHAT to do
- there is a low level KI to decide WHERE to do
- there is a AiStatistics class which collects data from the grids and is the interface for the high and low-level ki
- currently every player which is no human player will be one special computer player: RomanWhatToDoAi -> later it could be possible to create different AIs which can be chosen before game start. e.g. easy, medium and hard player, rusher, economist
- NetworkTimer can now notify listeners. This was chosen instead of sending AiExecutionTasks around the network because this was heavy IO and reduced the game performance
- Ai is executed only at the host computer
- I oriented at the YAGNI principle and implemented O(n²) first and take some measurements with jvisualvm and after that I implemented O(n*log n) only for the hot spots. The performance is ok at my mac. I don’t have experience with android. please give me feedback if there is an issue, than I keep optimizing (which will increase the complexibility of the code)

# What I noticed:
- running the AI helps to detect issues in the balance:
- insufficient number of solders or goods for map creation
- maybe to long time to produce goods (e.g corn)
- toolsmith: hammer and shovels should not be produced when other tools are needed. (e.g. the scythe is more imported because it leads to food which leads to iron ore and coal which leads to much more tools)
- no goods (especially gold) are not stored in stocks

Feedback is strong welcome!


.gitignore
```diff
@@ -2,6 +2,8 @@ bin
 gen
 target
 logs
+.idea
```
krichter722: IDE specific files (except eclipse because it's the official build platform) should be handled in `.git/info/exclude` because the number of IDEs and editor is endless and thus `.gitignore` might have an endless length one day.

krichter722:
When I play on the "big" map I hear building noise, but no buildings are planned by computerplayers.

experienced with v0.1.2-alpha-409-ga15c623

codingberlin:
@krichter722 the feature computerplayer is not part of the v0.1.2-alpha release.

krichter722:
> @krichter722 the feature computerplayer is not part of the v0.1.2-alpha release.

This is the output of `git describe --tags`. It refers to the 409th commit behind the v1.0.2-alpha tag (see `man git describe` for more details).

.gitignore
codingberlin: I didn't know ```.git/info/exclude``` so thx for the info.

I fixed this.

codingberlin:
> This is the output of `git describe --tags`. It refers to the 409th commit behind the v1.0.2-alpha tag (see `man git describe` for more details).
Thx for the info.

I was able to reproduce the described issue.
However with the current commit ( v0.1.2-alpha-550-gaea1344 ) the computer players places new buildings (towers, living houses, wine growers and storages ... and the only-one-tower players starts with living houses and lumber jacks)

andreas-eberle:
I tested the computer player a bit. The results are already really great! However, the repeated performance break-downs are sometimes quite severe... I guess this is due to the recalculation of the Statistics, isn't it? I will have a look to reduce the performance requirements by using more exsiting values or already calculating them during the normal game.

codingberlin:
One small improvement I added in the other branch enhancement_computerplayer_fights:
https://github.com/jsettlers/settlers-remake/blob/enhancement_computerplayer_fights/jsettlers.logic/src/jsettlers/ai/highlevel/AiStatistics.java#L272

This takes the already calculated material stats of the computer's main partition instead of recalculating it.
Furthermore I changed in the enhancement_computerplayer_fights branch the execution of computerplayer from every 3 to 10 seconds.


One thought I had was to split the computerplayers execution:
1) caculate AiStatistics
2) calculate 1. compi
3) calculate 2. compi
5) calculate n. compi
This would be possible if we do not schedule AiExecutor every 10 seconds but

AiStatistics, 1. compi, 2. compi, n. compi every 10 seconds with an increasing offset.
We then would have on the one hand the same execution time in sum but the gameloop would become time to do other stuff within the computercalculations.
What do you think about that?

codingberlin:
But of course you are right. If the statistics doesn't need to be recalculated every x secons but would be updated on demand (when building, when cutting trees/stones, when moving) ... this would boost the performance a lot.

michaelzangl:
You can try to asynchronize the computer player. This would allow you to do more difficult work without having to worry about the game loop. The current problem is that it still needs to be in sync for all players.

You can achieve  this by splitting up the computerplayer work in (small) parts that need to be synchronized with the game and parts that only rely on the data previously computed (statistics, ...).

Every time you query the computer player (every 10 seconds), you start with those tasks and and execute one every game tick. In the future, you can run the asynchronous tasks in a parallel thread and make that thread sync with the game after a given amount of ticks (e.g. 5 ticks).

michaelzangl:
One more thing: try analyzing your code, e.g. using visualvm. It is a bit difficult to find the right settings for it, but you could e.g. use profiling and set your computerplayer class as start class.

codingberlin:
The reason why I coupled the computerplayer with the game loop was the game speed. If you press F12, the computerplayer needs to be executed more often than with normal game speed.

At the velocity conference they told us, that it is important to set performance goals before you start to optimize the speed to know when you have finished.

So how many ms the computerplayer is allowed to use for one tick?

codingberlin:
As the computerplayer's performance is fast now, can this branch be merged then?

andreas-eberle:
Performance does seem ok most of the times now. => Merging this to master.

PR #153, opened 2015-09-12
Enhancement of mines, implementation of productivity calculation (fixes #151)

### LINKED ISSUE #151: Implement calculation of mine productivity
As of cb18427708f6ba3344463a2202b5287da2e872d2, the UI can now display the productivity of mines. This must now be calculated in the game logic and also be used for the mining action.

andreas-eberle:
Still to do: Calculation if the mining action was successful or not should be simmilar to the original game as described in #151.


andreas-eberle:
@codingberlin: Would you like to recheck the new behavior of the mines (in the game)? Do you see any problems? A mine should now be able to produce at max 1050 of its resource.

codingberlin:
@andreas-eberle In the original game I use a market and donkeys to do my test games.
Currently I can't do this with the remake so I can't test all I would like to test.

What I did so far is to build a coal and an iron mine and the productivity display felt well.
The miner took 10 baskets without food and took different amounts of baskets according to the preferred food.

Besides I build an iron mine a rock with less iron and some other resources. The productivity display felt well, too.

So in my opinion everything is fine.

PR #160, opened 2015-09-19
Fix NullPointerException when pathing

codingberlin:
All code examples are in file Movable.java
In the following code a NullPointerException can happen:
```
if (moveToRequest != null) {
	switch (state) {
	case PATHING:
		// if we're currently pathing, stop former pathing and calculate a new path
		setState(EMovableState.DOING_NOTHING);
		this.movableAction = EAction.NO_ACTION;
		this.path = null;
	case DOING_NOTHING:
		ShortPoint2D oldTargetPos = path != null ? path.getTargetPos() : null;
		ShortPoint2D oldPos = position;
		boolean foundPath = goToPos(moveToRequest); // progress is reset in here
		moveToRequest = null;
		if (foundPath) {
			this.strategy.moveToPathSet(oldPos, oldTargetPos, path.getTargetPos());
```
At ```path.getTargetPos()``` in the last line the NullPointerException occurs when ```goToPos(moveToRequest);```in the 11th line return true although the path was reset to null in:
```
final boolean goToPos(ShortPoint2D targetPos) {
	assert state == EMovableState.DOING_NOTHING : "can't do goToPos() if state isn't DOING_NOTHING. curr state: " + state;
	Path path = grid.calculatePathTo(this, targetPos);
	if (path == null) {
		return false;
	} else {
		followPath(path);
		return path != null;
```
```return path != null;``` returns true although ```followPath(path);```in the last line resets the path to null by
```
private void followPath(Path path) {
	this.path = path;
	setState(EMovableState.PATHING);
	this.movableAction = EAction.NO_ACTION;
	pathingAction();
```
```pathingAction();``` in the last line  
```
private void pathingAction() {
	if (!path.hasNextStep() || !strategy.checkPathStepPreconditions(path.getTargetPos(), path.getStep())) {
		// if path is finished, or canceled by strategy return from here
		setState(EMovableState.DOING_NOTHING);
		movableAction = EAction.NO_ACTION;
		path = null;
```
```path = null;``` resets path to null.

Checking for ```return this.path != null;``` instead of ```return path != null;``` will prevent the NullPointerException.

As soon as I understood the code this can happen when you send a movable to a position the movable is already.

### DIFF
--- a/jsettlers.logic/src/jsettlers/logic/movable/Movable.java
+++ b/jsettlers.logic/src/jsettlers/logic/movable/Movable.java
@@ -558,7 +558,7 @@ public final class Movable implements IScheduledTimerable, IPathCalculatable, II
 			return false;
 		} else {
 			followPath(path);
-			return path != null;
+			return this.path != null;
 		}
 	}
 


andreas-eberle:
Wow, many thanks for finding this!

Issue #162, opened 2015-09-20
adapt fight system to original settlers and add KK (combat strength)


codingberlin:
Soldiers have a combat strengh which increases by gold and is at minimum 100% within the own land.
Some years ago I made a test with the original game: https://tips.siedler3.net/page.php?lan=de&article=gold (german)

The soldiers have different hit points, strength and range.
Important is the range of 2 of pikemen. They are good in tight fights because they can attack from the second line but they are weak in wide areas due to their hitpoints and strength.

# strength
|soldier|hitpoints|strength|range|hit every x seconds|
|---------|-----------|-----------|-------|-------|
|swordsman lv1|100|10|1|0.7|
|swordsman lv2|120|14|1|0.7|
|swordsman lv3|150|20|1|0.7|
|pikeman lv1|200|4|1-2|0.7|
|pikeman lv2|240|5|1-2|0.7|
|pikeman lv3|300|6|1-2|0.7|
|bowman lv1|100|5|5 - 15 (25 from tower)|1.4|
|bowman lv2|120|6|5 - 15 (25 from tower)|1.4|
|bowman lv3|150|7|5 - 15 (25 from tower)|1.4|
|priest|80||
|pioneer|20||
|geologist|20||
|thief|20||
The list is based of: https://www.siedler3.net/cb3/verhaeltnis-von-kampfkraft-zu-level-der-soldatenp26794/#real26794 and http://www.siedlercommunity.de/c-Siedler-3-Hitpoints-s81.html and the measurements in the comments below.

# speed
|movable|speed [seconds per 100 tiles]|
|---------|-----------|
|priest|70|
|pioneer|60|
|geologist|60|
|bearer|60|
|thief|60|
|bowman|60|
|pikeman|50|
|swordsman|45|

IridiumOxide:
They also have different movement and attack speed. I've seen a different table: http://www.siedlercommunity.de/c-Siedler-3-Hitpoints-s81.html , but I don't know if the values are right, as hitpoints and strength seem to be wrong in this one. That site also states that Amazon soldiers have 20% less strength.

codingberlin:
I found that table, too.
It is correct, that the Amazon soldiers have 80% hit points. I left that out because currently the remake has only Roman.
I choose the table mentioned above because it is out of a manual and fits better my experience with the game.

codingberlin:
Regarding speed and attack frequency: @andreas-eberle is it possible to influence that with the current code?

[...]

PR #178, opened 2015-09-28
Feature combat strength

codingberlin:
This feature is part of #162.


codingberlin:
Thx for all your help.
I think we are ready to create a pull-request.
Please take a look at the performance again at BigMap where the LooserGeneral starts to fight the combat Strength is calculated for every hit any soldier does.

Should we calculate the combat strength for each player every 1000 ms and cache it?

andreas-eberle:
I changed the code to only update the combat strength when the gold amount is chaged.

codingberlin:
In my opinion everything is fine to merge now. Your opinion?

PR #181, opened 2015-09-29
Enhancement hitpoints, strength and ranges

codingberlin:
This is part of #162

- I introduced hitpoints and strength to the soldiers, pioneers, priests, geologists, thefts according to the table in #162.
- pikemen have a range of 2 now.
- fighting soldiers do not react on move order (like in the original game) (later we could introduce CTRL+rightclick to force moving and ignoring enemies as counterpart. like in the original game, too)


jsettlers.graphics/src/jsettlers/graphics/map/MapContent.java
```diff
@@ -605,7 +605,7 @@ private static Action getActionForKeyboard(String keyCode) {
 			return new Action(EActionType.SPEED_TOGGLE_PAUSE);
 		} else if ("BACK".equalsIgnoreCase(keyCode)) {
 			return new Action(EActionType.BACK);
-		} else if ("+".equals(keyCode)) {
+		} else if ("]".equals(keyCode)) {
```
andreas-eberle: Did you change this by intention?

codingberlin: Oops this should be on my "do not commit" part in idea :-)
This is another topic: at my mac the keycodes are different. The + at my keyboard is ] as keycode. I changed that (only for me) to be able to increase the speed.

codingberlin: I fixed this

codingberlin:
I noticed that the minimum distance for bowmen is missing. I will implement that.

codingberlin:
This pull request is still not ready yet.
On the one hand I add a minimum range to the bowmen but on the other hand this leads to strange behavior: If a bowmen likes to shoot an enemy which is nearer then the minimum range but there are other enemies within the range, the bowman will not shoot at any enemy.
If an enemy is nearer than the minimum range, the bowman will run directly into the sword of the enemy instead of running away.

So there is still something to do.

codingberlin:
I add the code so that a bowman escapes from an enemy within its minimum range instead of walking to him. But I feel a bit uncomfortable within the SoldierStrategy code and don't really know if the new code introduced some side effects.
@andreas-eberle, @michaelzangl please review it.

andreas-eberle:
The current implementation does not work with bowmen in towers, as they loose track of soldiers if all soldiers in the max search distance are in the minimum search distance. 
This causes the bowmen to thinks there is no enemy any more and therefore it changes its state to non-attacking. I'm trying to change it to fix this issue.

codingberlin:
Thanks for review and changing the code. The soldiers behave more fluent now. The disable-ai flag was a very good idea.

PR #184, opened 2015-09-30
Feature big temple

codingberlin:
- you can build a big temple
- the big temple spawns priests every 10 minutes
- you get 8 Mana when you finish a big temple until minimum one big temple is destroyed (to prevent building and crashing big temples while getting back the material to build new big temples until you have level3 soldiers)


codingberlin:
I found out, that the SpawnBuildings have Bug: When a settler is spawned but there is no space the settler is broken which leads to a settlers jam.
You can watch this when you let the KI play mountain lake to 1:10 then there are multiple big living houses where you can find a freezer settler at the door. After ca 2:00 this leads to a settlers jam after the big living house was destroyed in greens land near the spreading river.

Has someone an idea why this happen?

codingberlin:
I noticed that the second spawned bearer from the big living house mentioned above is broken because the first was not able to leave the door because he is blocked by a builder.

Bit this should be handled by the following check, shouldn't it?
```
Movable movableAtDoor = super.getGrid().getMovable(super.getDoor());

		if (movableAtDoor == null) {
			movableAtDoor = new Movable(super.getGrid().getMovableGrid(), getMovableType(), getDoor(), super.getPlayer());
```

jsettlers.logic/src/jsettlers/logic/buildings/spawn/BigTemple.java
```diff
@@ -0,0 +1,84 @@
...
+	@Override
+	protected int constructionFinishedEvent() {
+		getPlayer().getManaInformation().increaseManaByBigTemple();
+		iIncreasedMana = true;
+		return super.constructionFinishedEvent();
+	}
+
+	@Override
+	protected void killedEvent() {
+		if (iIncreasedMana) {
```
andreas-eberle: You can replace the iIncreasedMana field by checking super.isConstructionFinished().

jsettlers.logic/src/jsettlers/logic/buildings/spawn/SpawnBuilding.java
```diff
@@ -45,11 +44,11 @@ public final EMapObjectType getFlagType() {
 
 	@Override
 	protected int constructionFinishedEvent() {
-		return PRODUCE_PERIOD;
+		return subTimerEvent();
```
andreas-eberle: Maybe we should leve the 2 seconds period here. By this we ensure the first movable is spawned quickly after the building is finished, but after all builders really finished there work. I know this is a bit of a hack to compensate the non-ideal implementation of the builders but it could also look nice ;)

jsettlers.logic/src/jsettlers/logic/buildings/spawn/BigTemple.java
codingberlin: I thought the same but as I implemented it, the isConstructionFinfished() returned true when I killed the building although it was in construction phase when I killed it.

codingberlin:
Did I made something wrong or why there are only north-east builders and not north-west builders at the temple, too?

michaelzangl:
@codingberlin Builders cannot walk on blocked ground. Check that all builder positions are protected but not blocked.

codingberlin:
@michaelzangl thx for the hint. This was the problem. I fixed it.
In my opinion the code is ready to be merged now.

PR #197, opened 2015-10-10
Feature more buildings

michaelzangl:
This contains:
- Implementation of donkey farm
- (test) Market building + the ability to move materials there
- Stock building
- GUI for those.

There are still some bugs. The main bug is with the priorities of the stacks, this needs to be changed so that they are restored after re-scheduling a stack and that bearers don't drop materials when the priority is increased.


jsettlers.logic/src/jsettlers/logic/stack/RequestStack.java
```diff
@@ -82,6 +82,15 @@ public RequestStack(IRequestsStackGrid grid, ShortPoint2D position, EMaterialTyp
 		super.updatePriority(priority);
 	}
 
+	/**
+	 * Re-adds this request if it is not in any queue.
+	 */
+	protected final void reschedule() {
+		if (!isInQueue()) {
+			grid.request(materialType, this);
```
michaelzangl: This always requests using the default priority.

michaelzangl:
Fixes #214

Issue #201, opened 2015-10-12
AI is not working after loading a savegame


andreas-eberle:
#### Steps to reproduce
* Start a new singleplayer game with AI players.
* save the game
* quit the game
* load the savegame.

=> No AI players are active.

Issue #140, opened 2015-09-03
Implement new multiplayer join screen / singleplayer player selection screen


andreas-eberle:
### Current Situation
* There is no player selection screen for the singleplayer games. When starting a singleplayer game, the game directly starts without a chance to select the number of players / which players should be active / which start position (i.e. player number) the player wants to have.
* When joining or opening a new multiplayer match, there is a simple but buggy and ungly join screen. However, its functions are limited.

### Target Situation
A good join screen for both, single and multiplayer games. Possibly these screens look the same, as a singleplayer game is only a multiplayer game without other human players. 

#### Important Features
* Selection of player position (playerId)
* With the new AI pull request #139, the screen should also allow to add AI players to single and multiplayer games.
* Display a preview of the selected map with the player's start locations
* Multiplayer screen needs to have a ready button for all human players
* Multiplayer screen should display a chat window with the possibility to enter messages send to all match participants. The server should already support chat messages. 
* The chat window might also be used to display error messages (e.g. "Cannot start, because not every player is ready.")
* It should be possible to give the match a name, which will be displayed in the list of open network matches.

### Way To Go
I'm open for discussion on this topic. In fact, I opened this ticket as kind of a "master ticket", which will spawn new, smaller sub-tickets containing work packages. Feel free to comment if you want to work on a specific part. E.g. @adamjryan: you did a lot of UI work already, maybe, you want to develop the screen?

Certainly, we also need to do some work on the logic side here. I can implement such parts as required.

The Android implementation is a bit further developed. It could be used as an inspiration but should also be improved in the long term.

adamjryan:
I'd be happy to give this a go. Are you looking to reuse the original game graphics or would it be preferable to reduce the depedency on the proprietary stuff?

andreas-eberle:
The graphics of the original are very limited and therefore the possibilities to design this would also be limited. That's why I think we should/could use new graphics. There is already support for using own graphics files. @michaelzangl: Can you give more details on this (maybe write a wiki page?). Maybe you can "draw a picture" to give a first idea of the result and have a basis for discussions?

andreas-eberle:
I just had a discussion with @michaelzangl. We came to the conclusion, that we should not continue building the menu with our own OpenGL-based UI elements. Instead, we plan to create the main menu as SwingUI and skin that. @michaelzangl will implement a simple SwingUI version of the main menu.

adamjryan:
[...]

PR #207, opened 2015-10-19
feature computer difficulties

codingberlin:
To support issue #140 I introduce computer difficulties so that @adamjryan can make the following computer difficulties choosable by multiplayer join screen/singleplayer selection screen:

- **Very Easy** *LooserEconomyMinister & LooserGeneral*
- **Easy** *WinnerEconomyMinister & LooserGeneral*
- **Hard** *LooserEconomyMinister & WinnerGeneral*
- **Very Hard** *WinnerEconomyMinister & WinnerGeneral*

**WinnerEconomyMinister** is the high optimized computer player as you know it
**LooserEconomyMinister** builds less building material economy, less mana, late towers, inefficient weapons industry and causes sometimes lack of soldiers
**WinnerGeneral** is not finished yet because of missing features in the game. (Stop to recruit, store weapons, determine which weapons should be produced) (I add the WinnerGeneral in this early state to point to the idea to have different generals)
**LooserGeneral** produces inefficient mix of different troops, does not wait for level3, attacks only with the same number of soldiers as the enemy target has, so that a loose is very probable, because of the combat strength.

To use all 4 difficulties for a start the difficulty is chosen by round robin and depends on the playerId. Of course this will be removed, when the difficulty can be chosen by the UI later.

### DIFF
--- a/jsettlers.common/src/jsettlers/common/ai/EWhatToDoAiType.java
+++ b/jsettlers.common/src/jsettlers/common/ai/EWhatToDoAiType.java
@@ -3,17 +3,17 @@ package jsettlers.common.ai;
 /**
  * @author codingberlin
  */
-public enum WhatToDoAiType {
+public enum EWhatToDoAiType {
 	ROMAN_VERY_EASY,
 	ROMAN_EASY,
 	ROMAN_HARD,
 	ROMAN_VERY_HARD;
 
 
-	public static final WhatToDoAiType[] values = WhatToDoAiType.values();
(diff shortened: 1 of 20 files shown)


codingberlin:
This also fixes #201.
The problem was, that when saving the game the maxPlayerNumber = 1 was stored and not the correct one. So after the fix all new created save games will got an AI player (except the 0. player, this is the human)
After #140 is implemented, this should be reviewed again. Maybe then the 0. player is not the human but another id is.

Issue #215, opened 2015-10-26
Cannot drag-select units on high terrain


abledbody:
When units are standing on high terrain they are not selectable with mouse dragging.

andreas-eberle:
I also experienced this issue in the game. I think it is not depending on the terrain type, but on the height of the terrain. 

@michaelzangl: I think we already had this issue a long time ago. Maybe it was reintroduced?

justangel:
I have a similar problem

PR #217, opened 2015-10-27
Map Settler Craters

abledbody:
A little map I made consisting of 3 interconnected craters, surrounded by impassible snowy mountains.

Also an excellent demonstration of #215.

### DIFF
--- a/jsettlers.common/resources/localization/labels_en.properties
+++ b/jsettlers.common/resources/localization/labels_en.properties
@@ -13,128 +13,129 @@
 # DEALINGS IN THE SOFTWARE.
 ###############################################################################
 # buildings
-building_FORESTER    = forester
-building_LUMBERJACK  = lumberjack
-building_STONECUTTER = stonecutter
-building_SAWMILL     = sawmill
+building_FORESTER    = Forester's Hut
+building_LUMBERJACK  = Woodcutter's Hut
+building_STONECUTTER = Stonecutter's Hut
+building_SAWMILL     = Sawmill
+building_STOCK       = Storage Area
 
-building_COALMINE   = coal mine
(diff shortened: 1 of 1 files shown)


andreas-eberle:
Many thanks for building this map. It's great to have new maps for the players and for testing.

However, I think the current version has very small amounts of stone per player. Is this intended? Maybe we should increase the amount of stone.

abledbody:
@andreas-eberle I've been able to play the map without running out of stone, I don't think most players will before filling up their crater with buildings. I can up the stone count a bit if you think it needs it, but I think it might cramp the player in a little.

codingberlin:
Hey very good work!
This map helped to figure out some issues with the computerplayer.

(It would be wonderfully, if you add some small sea with fish for the goldmines.)

codingberlin:
By the way: I am such a player which fills the crater and runs out of stone by building a lot of living houses to produce mass of soldiers.

abledbody:
Oh, by the way, @codingberlin, Residences are only to keep Carriers happy, if you make a large residence and then make 50 soldiers you can destroy it to save some space.

codingberlin:
That's true, but you only get half of the material back thus you run out off stone sometime. But maybe low stone amout is one key point of the map which forces you to rush or attack in the midgame if you want to win.

However I think your map rocks:)

abledbody:
Just in case you guys missed it, I updated the map.

codingberlin:
+1

I tested the map again. From my perspective, it can be merged now.

abledbody:
@codingberlin Sweet.

andreas-eberle:
@abledbody: Thanks for the map!

PR #221, opened 2015-10-28
Bugfix: Checksum calc of original maps

tomsoftware:


codingberlin:
Big Thanks for that!

There are some java naming issues (e.g. upper and lower case) But this pull request goes into the feature_original_maps branch so that I merge it now and solve this issues later before we merge into master.

The next step is to generate MapList and MapFileHeader out of the original maps.
Unfortunately I do my boat driving license next Saturday so that I have to learn for it but after that I can keep on working on this.

Issue #210, opened 2015-10-23
Feature influence production rates of weapons


codingberlin:
I would like to implement the GUI and functionality to influence the production rates of weapons. (Because the computerplayer needs to influence it ;-) )

I would implement both: influence the ratios by increasing or decreasing the blue bars (Do we have somewhere such a bar already?) and production by prio (by increasing the numbers left to the blue bar)

I would focus on weapons but could implement tools as well, if we decide how to handle them in combination with the autoproduction of the toolsmith when tools are missing.
-> Should the autoproduction increase the numbers left to the blue bars so that the player is able to get feedback (and maybe influence it) what the toolsmith is producing?

codingberlin:
@michaelzangl while working on that issue I got two questions. Maybe you can help me?

1) how can an action be fired by BarFill when it is clicked? (e.g. like with ```new Button(Action action, ...)```)
2) how can I implement that the ToolsPanel refreshes the displayed information? (e.g. you add 10 bows and 2 were produced the panel shows 10 until I move a bit, than it refreshes)

Tanks in advanced!

michaelzangl:
1) Extend the BarFill. Override getAction(x, y) to construct a new action. Pass null as action to the constructor.
2) Use this class to help you (see usage in buildings menu for an example).
https://github.com/jsettlers/settlers-remake/blob/master/jsettlers.graphics/src/jsettlers/graphics/utils/UIUpdater.java

codingberlin:
Both helped me. Thx!

By the way I coupled now the automatical and manual Toolproduction.
If tools for workers are missing, they will be produced in their priority order. But now they were displayed in the ToolsPanel as well. If the player reduces them, they will be increased again (because the tool is missing). If no tools are missing, the ToolSmith produces the configured tools like the weapons: First the priority orders and then the ratios (blue bars)

codingberlin:
How I can prevent that the player can change the settings for other players, too?
Is there some code example for other partition based UI settings where the player is not able to change other players partition data?

michaelzangl:
No. Such a test is currently not implemented in all of the partition-wise settings. Simply ignore it for now, it is on the TODO-List ;-)

andreas-eberle:
[...]

PR #222, opened 2015-10-28
Feature influence weaponproduction

codingberlin:
See #210

What you can do now:
- Set up to 20 Bow/Sword/Spear productions
- Set ratio of Bow/Sword/Spear with blue bars

The weapon smiths will produce what you defined in the given ratio.
If you add 10 bows and 20 swords, bows and swords will be produces in 1:2 ratio until the prios were produces.

If you click on the tools prio buttons or bars, nothing happens.

The WinnerGeneral will produce some swords to occupy towers, 20 spears to defeat rushs and then only bows.


codingberlin:
Ich schau mir die kommenden Tage an, warum das buildSwing Target nicht mehr geht.

codingberlin:
(Sorry for writing the last comment in German. I even didn't notice that I did it :P)

The builds are green now so for my perspective this can be merged.

andreas-eberle:
@codingberlin: There seems to be a bug in the UI here. The tool tip message of the weapon production bars should display the percentage of swords, bows and spears that should be produced by the weapon smiths. At the moment, it only displays the percentage of how much the ui bar is filled.

codingberlin:
@andreas-eberle I fixed that now.
e.g. if you fill the bars swords 100% and bows 100% the bar descriptions show 50% now.

PR #227, opened 2015-11-04
Polish localization

IridiumOxide:
I've created the Polish localization file and tested it ingame - looks like everything is all right so far.


michaelzangl:
Thanks

PR #229, opened 2015-11-08
Feature original maps: import Buildings

tomsoftware:
Implementation of original maps import is complete ( 95 % )!!

Add more features to the "original reader"
+ read Settlers
+ read Buildings
+ read Resources
+ read Stacks
+ read Quest & Tip Text
+ some fixes

Known Bugs:
- Missing some Settlers (EMovableType) / Objects (MapObject) definitions in Lookup-Tables @ OriginalMapFileDataStructs.java
- Team Config is not loaded from Map-File
- Game final conditions are not loaded
- Map-Preview Image is not loaded (its just a 16 Bit image)
- ~~Border is not shown (but it is loaded to getBlockedPartition() )... this might be a Bug in the Main "Remake" Version... see getBlockedPartition @ OriginalMapFileContent.java~~
- ~~Remake MapEditor need to be fixed?~~


andreas-eberle:
@tomsoftware: First of all: many thanks for your assistance!
Regarding the blockedPartition: it has nothing to do with the borders of the players. The blocked partition are an additional data structure used to optimize path finding. They need to be calculated after setting up the landscape (like the map editor does it in https://github.com/jsettlers/settlers-remake/blob/master/jsettlers.mapcreator/src/jsettlers/mapcreator/data/MapData.java#L728). For this, the method should be extracted into the jsettlers.logic code. 

@codingberlin: I could have a look at this when this code is in your branch.

tomsoftware:
blocked partition is done (now the in-game-borders are shown :-) )... but I dit not move it to the jsettlers.logic code.

@andreas-eberle: what value range is used for 
byte getResourceAmount(short x, short y) 
Function return value? In the map file it is a 4-Bit value [0..15] and now I scale it up to [0..127].

codingberlin:
@tomsoftware just "Voll Gut!" ;)

I tested with Italy.map where you have some towers, settlers and goods. -> Worked
I tested with A640-4-River.map -> Worked ... and I saw some issues where I can optimize the AI. -> Perfect ;)

tomsoftware:
padding: sorry... I forgot to remove! It was just for my testing. (see new pull request with fix)

Issue #211, opened 2015-10-23
Read original maps


codingberlin:
It should be possible to read the original game maps in the remake.

codingberlin:
@tomsoftware would you mind to help us here with your knowlede, you collected in your repository? https://github.com/tomsoftware/sied3

codingberlin:
@andreas-eberle @michaelzangl What do you think about transforming the original maps when jsettlers is started once (when the maps weren't transformed yet) into the new map format and store them in the maps dir?

michaelzangl:
This might prove to be difficult. We use the same graphics, but our game grid has many differences. We can attempt to translate the maps but we might have problems in translating the images.

Our main goal should be to support an intuitive, good map editor and create our own set of maps.

So far, supporting those files is not planed, although the file format should not make any difficulties.

codingberlin:
@michaelzangl When we know what the bits and bytes of the original map format means (e.g. Water Level 1) we can transform it into the new map format (Water Level 1) which can be read by the remake.

andreas-eberle:
I think the best way would be to implement a loader for the original maps which implements the jsettlers.common.map.IMapData interface. This interface defines all the information required for creating a new MainGrid (i.e. for loading the map to start a game) and can also be used to load the map into the MapEditor.

I would not transform the maps into our map format. This is because 
* we would need the loader implementing the IMapData interface anyways (at least it must supply the same information). So converting them is actually one step more that just reading them on demand.
* we probably cannot transform the maps and ship them with the game (as we might not have the right to do so)
* if we provide a stable way to read the maps on demand, every user could just use his favorite maps which will include maps we don't know.
* by making it possible to directly start working on these maps with our MapEditor, people can develop there maps further with our tools and then easily create maps with our map format.

codingberlin:
I agree. Currently you will find prototype code in the branch I created: https://github.com/jsettlers/settlers-remake/blob/feature_original_maps/jsettlers.logic/src/jsettlers/logic/map/OriginalMapLoader.java
Because first I have to understand how the original map format works.
When I (or we) got it, then I will refactor the code so that it fits into the game as a MapLoader which gives IMapData and MapFileHeader (which is needed by the map lists, isn't it?)

tomsoftware:
What's missing?

As you can see:
https://github.com/jsettlers/settlers-remake/pull/230#event-461021927
the feature is working.

Known issues:
- Map creater can not save the maps
[...]

PR #235, opened 2015-11-11
Original Map Loader Fixes #2

### LINKED ISSUE #2: Settlers folder must be configurable on local developer machine
Furthermore, the paths currently configured in the global properties should be removed or replaced with the settlers standard installation path.

tomsoftware:
- Fix: Maploader saves the wrong Map Type if original was loaded.
- change map-heigh scale to 0.3 (see issues #211)
- Add: Quest-Text to original MapLoader "getDescription()"
- Fix: disable Quest-Text output


codingberlin:
You reduced the scale of the heights because you was not able to build a building in the remake where you are able to build it in the original game.
But not only the heights affect if you can build a building. The blocking tiles of the building and the algorithm in the remake which determines the possibility to flattern influences if you can build a building or not.

Comparing the River map loaded in the original game ...
original

... and loaded in the remake with factor 0.3 ...
remake

... and loaded in the remake with factor 0.56 (127f / 225f) ...
factor05

... I would say factor 0.3 is too flat.

Until now it was not possible to compare the behavior constructions between the two games on the same map which leads to the conclusion that the other factors (flattern algorithm, blocked tiles of buildings etc. cause that you can't build your building.

@andreas-eberle what is your optioning about that?
(by the way: It seems to me that the original game has more or darker shadow in the landscape)

tomsoftware:
@codingberlin with the building-dot-preview you can better compare the ("gefühlte") high in the games (see issues #211 [factor 0.5] and https://github.com/jsettlers/settlers-remake/pull/230 [factor 0.3 and original])... but may you are right: the real factor is somewhere in between 0.3 and 0.4 but 0.3 is not bad ;-)

There are match other differences in the remake e.g.:
- shadow of the hills
- you can’t build on the left-front-side of a hill
- the territory claimed by a tower is too big in the remake
- no decoration elements
but this is not a loader problem and need to discuss later in a other issue :-)

codingberlin:
@tomsoftware I talked with @andreas-eberle about the building-dot-preview. He confirmed that not only the heights influence if you can build a building in the remake and that we should adjust the other factors in order to get a 1:1 translation of the heights between the original maps and the remake.

Thus I adjusted with commit 928339d7618c565d41dc168c37e82c953465be24 the ```CONSTRUCTION_MARK_SCALE_FACTOR``` which influences when to build a building is allowed.
I tested this with an own map loaded in the original game and the remake.
The construction marks does not fit exactly but better now. (This is because of the other mentioned factors e.g. the blocking tiles of a building which could differ between the original and the remake, too)

Could you verify that the new ```CONSTRUCTION_MARK_SCALE_FACTOR``` lets you build the lumberjack at your testmap running the remake with ```ORIGINAL_TO_REMAKE_HEIGHT_FACTOR = 127f / 225f``` and tell me if the new ```CONSTRUCTION_MARK_SCALE_FACTOR``` is ok or if I need to adjust it a bit more?

Regarding this pull request, please revert the changes:
```
//- Heigh of original maps are 0..225 and of remake 0..127 -> x = 127f / 225f;
//-  but remake allows much higher mountains -> x = 0.3
private final static float ORIGINAL_TO_REMAKE_HEIGHT_FACTOR = 0.3f;
```
Then I can merge the fixes and features of this pull request. (Thx for implementing them :)

Thx in advanced.

codingberlin:
Thx

PR #239, opened 2015-11-14
make shadow darker

codingberlin:
This PR is to make the landscape more plastic by making the shadow darker.

The landscape before the PR:
bildschirmfoto 2015-11-14 um 09 30 47

The landscape in the original game:
original

The landscape after the PR:
bildschirmfoto 2015-11-14 um 09 32 05

With this PR the landscapes become more closer to the original but as you can see in the screenshot from the original game there the hills are not only dark but lighten as well.
I was not able to find out how to lighten the textures. even by setting the colors to 255 255 255 didn't lighten them, this just let the original landscape color as it is.


andreas-eberle:
@michaelzangl: Is there a way in our code to "lighten up" some parts of the textures?

michaelzangl:
No, currently I only darken the textures. The best thing would be to write a shader for this that does the brightening.

codingberlin:
Can this ne mergen then or is there something which speaks against it?

PR #242, opened 2015-11-15
Endgame statistik and unit test for AI

codingberlin:
Added Endgame statistic which is currently printed out in the console when the player stops the game.
It can be used and implemented further later when the "Game Finished" screen is available.

- Endgame statistic is used in new added AI test, which checks if any change in the code makes the ai weaker.
- Add tests to compare the different difficulties of the AI
- All AI tests checks if the runtime performance is influenced by code changes


codingberlin:
As you can see, 2 AI Tests are @Ignored. With the replay tool I found out that in the games unoccupiable towers exists.
I suggest: merge this pull request and when towers are fixed, Un-Ignore the tests.
The tests can be used to reproduce the bug while fixing it.

PR #243, opened 2015-11-15
Fix multiple ai towers at one position

codingberlin:
Observed on River map the AI player built multiple towers at the same spot instead of spreading the land.
This was caused by two issues:
1) The land around swamp was recognized as border land.
-> I fixed that by using ```PartitionsGrid.isDefaultPartition()``` instead of my own check.
2) The computerplayers tried to expand to the hills of their enemies.
-> I fixed that by not forcing the AI to expand in direction of coal if all resources are owned already. Now in this case the AI just expands in all directions by building towers most away from all other builded towers.


codingberlin:
While working on another issue I noticed that the algorithm lead to broad but thin land but such a land is difficult to defeat. Thats why I changed the algorithm again.

If all needed resources were found, the ai shuffles the decision 50:50 if a tower to spread the land around the center of mass of the player's land or to go for more gold/iron/coal with the next tower.

jsettlers.logic/src/jsettlers/ai/construction/BestMilitaryConstructionPositionFinder.java
```diff
@@ -173,7 +190,10 @@ public ShortPoint2D findBestConstructionPosition(AiStatistics aiStatistics, Abst
 				aiStatistics.getNearestResourcePointInDefaultPartitionFor(referencePoint, EResourceType.GOLDORE, Integer.MAX_VALUE) != null) {
 			importantResources.add(ImportantResource.GOLD);
 		}
-		if (importantResources.size() == 0) {
+		// 50 : 50 chance to spread the land or to go for more resources
+		if (importantResources.size() == 0 && RandomSingleton.getBoolean() == true) {
```
andreas-eberle: We cannot use the RandomSingleton in the AI. This is because the AI is only calculated in the game of one player. Therefore this player would have a differing use of RandomSingleton from the others (if bots are used in multiplayer games). 
However, we also cannot use Math.random(), as this would make the AIs behaviour interdeministic and therefore unreproduceable which would make debugging a pain. 

My proposal: Add an AI random generator only used by the AI and initialized with the same seed as the RandomSingleton (or the first random number of the random singleton). The important thing is that this seed is deterministic but not fixed.

andreas-eberle:
@codingberlin: Did you have a look at my changes? Are you ok with them / did you find any issues? If not, I would merge this to master.

codingberlin:
@andreas-eberle sorry I didn't notice you commits yet.
This was a full week for me. If it's ok for you I would watch them in the next days and then merge them?

andreas-eberle:
That's fine for me.

codingberlin:
@andreas-eberle Very good refactoring idea. Well done.

Issue #240, opened 2015-11-15
NullPointerException on startup


andreas-eberle:
Sebastian send me this by mail: 

I just wanted to try out your settlers 3 remake and got the following
Exception:

g:\Spiele\JSettlers>java -jar JSettlers.jar
Exception in thread "main" java.lang.NullPointerException
        at sun.awt.shell.Win32ShellFolder2.getFileSystemPath(Unknown
Source) at sun.awt.shell.Win32ShellFolder2.access$400(Unknown Source)
        at sun.awt.shell.Win32ShellFolder2$10.call(Unknown Source)
        at sun.awt.shell.Win32ShellFolder2$10.call(Unknown Source)
        at java.util.concurrent.FutureTask.run(Unknown Source)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown
Source) at java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown
Source) at
sun.awt.shell.Win32ShellFolderManager2$ComInvoker$3.run(Unknown Sourc e)
        at java.lang.Thread.run(Unknown Source)

andreas-eberle:
I requested further information about OS type/version, used game release and Java version.

andreas-eberle:
This might be caused by http://bugs.java.com/bugdatabase/view_bug.do?bug_id=6973685

UnitedHaxor:
Hi,
I've used https://github.com/jsettlers/settlers-remake/releases/download/v0.1.3-alpha/JSettlers_v0.1.3-alpha.zip

I'm using Windows 7 64-bit and Java8:

D:\Users\United>java -version
java version "1.8.0_51"
Java(TM) SE Runtime Environment (build 1.8.0_51-b16)
Java HotSpot(TM) 64-Bit Server VM (build 25.51-b03, mixed mode)

But it does not work with Java 7 too:
E:\Java\jre7\bin>java -version
java version "1.7.0_45"
Java(TM) SE Runtime Environment (build 1.7.0_45-b18)
Java HotSpot(TM) 64-Bit Server VM (build 24.45-b08, mixed mode)

E:\Java\jre7\bin>java -jar c:\JSettlers\JSettlers.jar
Exception in thread "main" java.lang.NullPointerException
        at sun.awt.shell.Win32ShellFolder2.getFileSystemPath(Unknown Source)
        ...
        at java.lang.Thread.run(Unknown Source)


[...]

PR #248, opened 2015-11-22
Change using of JFileChooser() to FileDialog() cause issue #240

tomsoftware:
- Change using of JFileChooser() to FileDialog() cause issue #240
- Known bugs: the nativ FileDialog() only can select a File not a Folder (the nativ DirectoryDialog would require the  import of org.eclipse.swt.*)
- some "/" to "File.Separator" fixes

### DIFF
--- a/jsettlers.graphics.swing/src/jsettlers/graphics/swing/resources/SwingResourceProvider.java
+++ b/jsettlers.graphics.swing/src/jsettlers/graphics/swing/resources/SwingResourceProvider.java
@@ -27,7 +27,7 @@ public class SwingResourceProvider implements IResourceProvider {
 	private final String resourcesFolder;
 
 	public SwingResourceProvider(File file) {
-		this.resourcesFolder = file.getPath() + file.separator;
+		this.resourcesFolder = file.getPath() + File.separator;
 	}
 
 	@Override
--- a/jsettlers.graphics.swing/src/jsettlers/graphics/swing/resources/SwingResourceProvider.java
+++ b/jsettlers.graphics.swing/src/jsettlers/graphics/swing/resources/SwingResourceProvider.java
@@ -27,7 +27,7 @@ public class SwingResourceProvider implements IResourceProvider {
 	private final String resourcesFolder;
 
 	public SwingResourceProvider(File file) {
-		this.resourcesFolder = file.getPath() + "/";
+		this.resourcesFolder = file.getPath() + file.separator;
 	}
 
 	@Override
--- a/jsettlers.main.swing/src/jsettlers/main/swing/SwingManagedJSettlers.java
+++ b/jsettlers.main.swing/src/jsettlers/main/swing/SwingManagedJSettlers.java
@@ -15,6 +15,7 @@
 package jsettlers.main.swing;
 
 import java.awt.Dimension;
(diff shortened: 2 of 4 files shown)


jsettlers.graphics.swing/src/jsettlers/graphics/swing/resources/SwingResourceProvider.java
```diff
@@ -27,7 +27,7 @@
 	private final String resourcesFolder;
 
 	public SwingResourceProvider(File file) {
-		this.resourcesFolder = file.getPath() + "/";
+		this.resourcesFolder = file.getPath() + file.separator;
```
andreas-eberle: Please use a static access (File.separator).

jsettlers.main.swing/src/jsettlers/main/swing/SwingManagedJSettlers.java
```diff
@@ -79,6 +81,44 @@ public static void main(String[] args) throws FileNotFoundException, IOException
...
+	private static String getSelectedFolderName() {
+		
+		//- Info: Don't use JFileChooser(), some Windows User have problems caused by a NullPointerException in FileDialog()
+		//-   see issue #240: https://github.com/jsettlers/settlers-remake/issues/240
+		
+		
+		//- see: http://stackoverflow.com/questions/1224714/how-can-i-make-a-java-filedialog-accept-directories-as-its-filetype-in-os-x
+		JFrame frame = new JFrame();
+		
+		System.setProperty("apple.awt.fileDialogForDirectories", "true");
```
andreas-eberle: apple.awt? I would really recommend not to use platform specific stuff. 

On Windows I'm unable to select a folder. When I select a file, I have to know that I have to select a file in the root folder of the Settlers 3 installation to make it work. However, I don't get this information from the dialog.

tomsoftware: # apple.awt : I know...this is the recommendation on Stackoverflow.
# Select folder on windows: yes, I know. The default Folder-Selection-common-Dialog is not available with Java (only with the org.eclipse.swt.* pack) so may we change the title of this dialog to "selet the file x.y"
# alternative:
1. stay with JFileChooser() and tell users with the bug to change the config by hand
2. use JFileChooser() with a fallback to FileDialog()
3. create ouer own directorychooser
4. use the (org.eclipse.swt.*) Dir File Selection

andreasb242: I have wrote a directory chooser in Swing, which looks similar to the Windows one.
But I have to remove some dependencies, then it can be copied, because I will not use the whole library...
Should I prepare it?

UnitedHaxor:
Hi,
in alpha-0.2 it doesn't work, too. Which config do I need to change / specify?

andreasb242:
This is a property file, you should use two backslash instead of one, this means settlers-folder=g:\\spiele\\s3 or use forward slashes (/), then you can use only one.

UnitedHaxor:
Thanks, that worked.

tomsoftware:
The pull request #299 solvs this mutch better... So I'd like to close this.

PR #256, opened 2015-12-27
Feature command line flags file

andreas-eberle:
This PR adds an optional options.prp file that can be used to set a command line options every time and without the usage of a console window (this is a bit more user friendly until we have a better UI).

Furthermore, a new command line option "fixed-ai-type" is introduced. It can be used to specify the AI type that should be used for all AI players (instead of having varying AI difficulties). 

@codingberlin: Can you have a reviewing look at the code?


codingberlin:
My first thought: An additional section in the Readme.md would help other people to
a) find the options.prp
b) understand what they can influence there

... now I will try it out and will give further feedback

andreas-eberle:
Your're right, we should add a section in the readme.

codingberlin:
I'm done with me review.

I would suggest instead of replacing the command line args by options.prp file, to mix both mechanism together with following order:
- defaults implemented in the game code
- is overwritten by options.prp
- is overwritten by command line args

With that mix I have the opportunity to set ```all-ai=true```in the options.prp because I as the AI developer need AI's everywhere ... but I can create one run-configuration in my IDE where I use the GUI to select my map or maybe a multiplayer game etc. and I can create another run-configuration where I can add a command-line argument for a specific map for which I fix the current bug.

PR #258, opened 2015-12-31
Improve editor gui

andreasb242:
* New file menu and toolbar, configured by a property file, same actions for Toolbar and Menu
   * Keyboard shortcuts also defined in a seperate property file
```INI
# Main menubar
menubar=file,edit,view,help

## Translations:
## Key menu.xxx is used as translation key
## "---" is a separator

menu.file=new,open,save,save-as,open-map-folder,---,play,---,export-image,---,quit
menu.edit=undo,redo
menu.view=zoom-in,zoom-out,zoom100,---,show-statistic,show-map-settings,show-tools,---,show-errors,goto-error
menu.help=manual,about

## Toolbar
toolbar=new,open,save,---,undo,redo,---,play,---,zoom-in,zoom-out,zoom100,---,show-statistic,show-map-settings,---,show-errors,goto-error,---,player-spinner
```
* Sidebar with Tabs for errors and tools, tools have icons (not all yet)
s3-4
   * Sidebar is resizeable
s3-5
* Open or create a new file from editor, start a new instance
* New start Dialog, open mask list formatted
   * Additional Last used tab added, which shows (max.) 10 last used maps
s3-1
s3-2
s3-3


andreasb242:
"Save as" and *Export as image" is not yet implemented.

andreasb242:
Preview is now displayed in the Open window... I coulden't find an example, and the documentation could be better ;-)
s3-10
Landscape type is now translated, and the color is displayed, as in the tool tree also
s3-11
Player selection now shows the player color, and the height tools have icons.
s3-12

And a lot of small changes...

jsettlers.mapcreator/src/jsettlers/mapcreator/main/DataTester.java
```diff
@@ -49,21 +52,48 @@
...
+	/**
+	 * Release all resources
+	 */
+	public void dispose() {
+		running = false;
+		synchronized (this) {
+			notifyAll();
+		}
+		try {
+			thread.join(100);
```
michaelzangl: Why the timeout? We should at least handle the case that the thread runs into a timeout.

michaelzangl:
Well done. Yes, the editor was a bit hacked. This makes the editor much more usable.

jsettlers.mapcreator/src/jsettlers/mapcreator/main/DataTester.java
andreasb242: Currently the editor is always quit with System.exit, because the re are a lot of threads which aren't stopped.
But you're right... It makes no sense as it is currently, i add a thread.stop() to kill after, OK?
```java
			thread.join(100);
			thread.stop();
```
Not the nice solution, but its currently anyway only at the end, and the thread should have enough time to quit.

andreasb242:
Thank you :-)
Only the shape type selection is not good, really bad usability...
I'll have to think about, how a good selection can look like, currently checkboxes and settings appear / disappear on selection tools...

jsettlers.mapcreator/src/jsettlers/mapcreator/main/DataTester.java
michaelzangl: thread.stop() is very bad. You could throw an exception if the join was not successful. Or simply remove the timeout: This allows users to file a bug report if the thread hangs and we can trace the source of the problem easier.

andreasb242: System.exit is also very bad...
This is not my code. I only add the dipose().

For the moment it's working.
Can we do later a refactoring, and use an ExecutorService service instead?
Then we don't have the problem. 
Then also the whole synchronisation stuff can be removed, if there a new check should be done, we create a new job, which is executed in the Threadpool (single threaded).

OK for you?

Off topic, but I hope you read this here:
@michaelzangl you already have 144 commits on this project, you may know it:
Was there already a discussion about error loggin and exception handling?

I added a default thread exception handler in the main of the Map Editor, which catches all unhandled exception, display a popup and print the stacktrace to stderr.

I also used this popup on all important catch blocks I found, before only the exception was logged, you clocked somewhere and nothing happend. Really annoying as user.

I would generally show an error if an exception happens which matters.

And I would prefer to use SLF4J, you have a lot more possibilities (turn off for performance, detailed logging to stdout on debugging etc.)

I didn't see any discussion about this topic...

michaelzangl: I have more than 1k commits, github is not counting all of them.

The DataTester was simply killed on exit because we know that we do not need it's data then. Stopping it by force may lead to corrupt DataTester data. I cannot guarantee that the rest of the code will work then.

Yes, there were thoughts about error logging but for the moment we just stuck to System.err.

We just don't have anything like a clear GUI exception handling. The GUI always attempts to catch them (if any could happen) or it simply crashes - in hope that users report the bugs they find.

Adding a good logging system would be a big improvement to the Project. #259

andreasb242: The Thread is not the only problem, the List with the errors, is handed over in the UI Thread, but not copied, so the The Checker thread and the UI Thread are working on the same list... Not good, I have to fix this...

andreasb242: I didn't see your commend, github didn't automatically refresh...
I'll fix the Datatester, and see if I can improve it a little bit with the new sidebar.
I'll extends my swing exception handler in the Map Editor. It can be later moved to use also for the game. Now I'm only working on the Map Editor, I don't know the Code enough good.
(And the Map Editor don't have to run on Android, makes it for the beginning a lot easyer...)

Logging is then may the next task...

michaelzangl:
Can you merge this branch? I did some enhancements on the error reporting.

https://github.com/jsettlers/settlers-remake/tree/andreasb242-improve-editor-gui

andreasb242:
Should be merged, but the automatically build is not starting any more?

andreasb242:
My local build is working. I changed the ant file, because I added JUnitTests for the Map Editor. Can somebody check whats wrong? I think all should run... :-(

andreasb242:
Build is working, but with disabled JUnitTest

@michaelzangl Can you have a look, the File "ValidationTest.java_" contains the JUnitTest for the MapCreator, which has to be added to the Ant Build path.
I couldn't get it build with ant...

andreasb242:
The most problems which are detected in the editor can also be automatically fixed.
The original Settler 3 editor blocks painting near the boards, not my prefered way, I like it much more like Eclipse: Let the user do all mistakes, but give hime the option to automatically fix it.

Only the first simple implementation is done, to take a screenshot :-)

s3-30
s3-31
s3-32

andreasb242:
Added icons for Shape types, display all types always, but disable the not supported types, makes the UI less flickering.

@michaelzangl 
I need some description / help: I would expect that I can draw a straight line with the line tool, but I couldn't figure out how?

s3-40

andreasb242:
Again some screenshots of my progress, and some description...

Open map with correct preview display, search within map name, description and ID
s3-50

Menu option to display resources always. If you select a resource tool or a tool to build a mine or a fishing house, the resources are also displayed (even if the checkbox is not checked)
s3-51

You can do a lot of non working stuff, e.g. if you draw gras over water with fishes, the fishes stay there.
You can place settlers out of the land of the owner... (you may would like to build a tower later?)
The errors are displayed at the right sidebar, errors are grouped, 10 settler which are next together will displayed as one error, instead of one error per settler...
s3-52

Now there is an auto fix option in the new "Error / Warning" menu, or you can press F9...
s3-54

And the errors are gone!
Not all errors can be fixed automatically, and in this case the fishes and settlers are simple deleted...
s3-55

andreas-eberle:
I just opened a new PR #272 as follow up of this PR. Therefore I will close this PR. All work will be continued in the new PR.

PR #273, opened 2016-01-09
Fixed wrong encoding of Umlaute

andreasb242:
And it should be even more efficient.
You cannot test it with this branch, but in the new OpenPanel the Description with "ö"s are now correct displayed.


andreasb242:
Obsolet, integrated into https://github.com/jsettlers/settlers-remake/pull/284#issuecomment-172231617

PR #288, opened 2016-01-17
Add copyright notice to new files

michaelzangl:
This fixes
- UTF8 issues for language files
- Add the default copyright header to every file created by default.


andreas-eberle:
Please add a section to the readme, explaining that we use this tooling to add the copyright notice in eclipse. Maybe others can add how to do it for other IDEs later.

michaelzangl:
I consider this a convenience setting that is not neccessary. "Normal" users can still copy their notes together.

This will not add the copyright notice to all files, it simply changes the default format for new files.

jsettlers.mapcreator/.settings/org.eclipse.core.resources.prefs
```diff
@@ -0,0 +1,4 @@
+eclipse.preferences.version=1
+encoding//src/jsettlers/mapcreator/localization/labels_de.properties=UTF-8
+encoding//src/jsettlers/mapcreator/localization/labels_en_US.properties=UTF-8
+encoding//src/jsettlers/mapcreator/localization/labels_ru.properties=UTF-8
```
andreas-eberle: Is it possible to specify that all files in /localization have to be UTF-8 encoded? This would make it save for the future. (Same with commons project).

andreas-eberle: It seems like this isn't possible, as Eclipse always interprets .properties files with their Java default encoding. It should be possible when we switch to another ending.

PR #294, opened 2016-01-19
Increase possible player count to 32.

andreasb242:
New Color array: The first 20 Colors are the original colors (Taken from Map Editor Screenshot)
The other 12 Colors are defined by me, and should be as unique as possible:
s3-colors
I increased the player limit to 32, which was the text in the Exception. The exception text is now taken from the property, which is also increased to 32.

There is now also a check, which checks all objects if they have a vilid player ID. If not, an error is displayed.

Sorry for the formatting change, I added some javadocs, nearly no changes, but I have the formatter activated in the save action, as written in the readme.


abledbody:
It looks a little hard to distinguish 21 from 4 and 24 from 15. I propose the alternatives R-163 G-197 B-3 and R-180 G-141 B-203 respectively.

andreasb242:
bildschirmfoto vom 2016-01-20 18 22 54

Your proposal 1 (P1) is nearly the same as The original Color 17.
The second is next to 20, but I think would be OK.

So I would replace 24 with P2?

For my tests I used a HTML file (not nice but easy to test):
[color2.html.zip](https://github.com/jsettlers/settlers-remake/files/97800/color2.html.zip)

It's not that easy to find enough colors, may because of this the original game was limited to 20 players ;-)

abledbody:
Oh, I see what you mean about P1. I don't know how I missed that. But I found a color that would work better, #CD0973 with the closest match being 31. So yes, the colors may look fine the way you have them now, but for distinguishing them better I do propose replacing 21 with P1 and 24 with P2. Although looking at it now P1 is about just as distinguishable from 31 as 21 is from 4, so it might not need to be changed.

Color pallets can be such a pain.

michaelzangl:
We can also simply add an option to change to enemy-colors. Then we only need three colors (me, enemy, friend)

andreasb242:
For me the second proposal of @abledbody seems to be OK.
Yoy even only run into this problem if you use 32 players, so I would call it a luxury problem ;-)

tomsoftware:
@michaelzangl : I like this idea... maybe the user can switch between this color schemes by clicking one button near the minimap.

andreasb242:
I changed Color 21 and 24.

bildschirmfoto vom 2016-01-23 16 49 12

andreas-eberle:
@andreasb242: Can you update the branch, so I can merge it?

andreasb242:
Done

PR #295, opened 2016-01-20
Bugfixes minor fixes for master and minimap color update

andreas-eberle:
Due to a recent change of @andreasb242, the minimap colors changed significantly (he tried to adapt it to the original game). I think the colors are to bright and we should choose our own minimap colors. Therefore I tried to change them a bit. Here is a comparison:

old
Our minimap before (please note: all water levels had the same color)

butti
@andreasb242's minimap colors

neu
Proposed minimap colors


michaelzangl:
I'd like to keep them darker.

The main reason for me to do so is to make players more visible (at least for few players). Compare the visibility of green in those.

I would also like to decrease the brightness of snow a bit to make the hill shade better. We should also increase the brightness of the river-blue and make hill shading adapt to the map size.

andreas-eberle:
@michaelzangl: Can you implement the adaption to the map size? I will have a look try to change colors following your suggestions.

PR #296, opened 2016-01-20
Fix debug colors

michaelzangl:
(includes eclipse path fixes)


andreas-eberle:
@michaelzangl: I extracted the actual fix for the debug colors to PR #305. Can you please rename this PR to what it really does change? Preferably with a meaningful description...

PR #300, opened 2016-01-23
Editor UI fixes

andreasb242:
This Pull request is based on the "Increase possible player count to 32" Pull request, and contains only 4 Commits:
* Preview image angle is now in the right angle, looks near the original one
* If map Names start with a space, it looks ugly, I fixed this in the view, It may is a Bug of the maploader which has also to be fixed there
* At the startup of the editor, it jumps to the center of the map (instead of the left bottom corner)
* Zooming is now possible direct after the start of the editor (I did a change in go.graphics.area.Area, Line 390, please check if this is OK!)


go.graphics/src/go/graphics/area/Area.java
```diff
@@ -396,6 +388,12 @@ public void handleEvent(GOEvent event) {
 				activeRegion.handleEvent(event);
 			}
 		} else if (event instanceof GOZoomEvent) {
+			if (activeRegion == null) {
+				// if there is no active region, set the first active,
+				// so the zoom is working in the Editor, even if you
+				// didn't press a mouse button
+				setActiveRegion(regionPositions.get(0).getRegion());
```
michaelzangl: Please check that regionPositions.size() > 0.

andreasb242:
>Please check that regionPositions.size() > 0.

Done

PR #304, opened 2016-01-25
Map editor warnings

andreasb242:
This pull request is based on my last one, and therefore contains also the changes of it.

Warning system implemented, now not only errors, also warnings are supported. The general idea behind: An error means the map is not playable, a warning is only an information, that you may missed something, but you can play the map. e.g. a map without stones can be played, if there all buildings already created in the map editor.

Some checks are now implemented. I'll may implement them later more efficient, but for now it's working.

bildschirmfoto vom 2016-01-25 21 12 43


andreas-eberle:
unbenannt

Sometimes you use "Kein(e)" and sometimes "Keine". Is there a case where you'd need the "(e)"?

andreasb242:
Kein(e) is used because I got the place holder from the existing Translation, and it's:
- Kein Gold
- Keine Kohle
...

I wrote some simple checks, there is a lot more which can be checked.
The important part of this Pull request is not this check, it's the interface for Warnings.

I think in the future we should share the game logic which analyses the map with the editor, e.g. Partition detection and check each partition for trees.

andreas-eberle:
I agree.

PR #299, opened 2016-01-23
New Settler 3 folder selection dialog (Startup of the Game)

andreasb242:
bildschirmfoto vom 2016-01-23 16 32 07

Tree folder selection, valid Settler folders are highlighted with a red arrow.
Help Button opens an URL, which points currently to wiki, I currently don't know where I should link it, but I would prefer a separate How-To (because I read the DEMO installation also works).

PR #306, opened 2016-02-28
New Settler 3 folder selection dialog (Startup of the Game)

andreas-eberle:
This is the follow up of #299.


jsettlers.main.swing/src/jsettlers/main/swing/foldertree/SelectSettlerFolderDialog.java
```diff
@@ -0,0 +1,304 @@
...
+
+	/**
+	 * The tree
+	 */
+	private JTree tree;
+
+	/**
+	 * Help URL: TODO Enter a valid help URL!
+	 */
+	private static final String HELP_URL = "https://github.com/jsettlers/settlers-remake/wiki";
```
andreas-eberle: Shouldn't we open the readme here? At least as long as we don't have any real wiki..

jsettlers.main.swing/src/jsettlers/main/swing/foldertree/FileTreeCellRenderer.java
```diff
@@ -0,0 +1,109 @@
...
+		public int getIconHeight() {
+			return 16;
+		}
+	};
+
+	/**
+	 * Constructor
+	 */
+	public FileTreeCellRenderer() {
+	}
```
andreas-eberle: In Java, the default constructor is always automatically implemented when you do not have another constructor. Therefore this constructor is not needed here.

jsettlers.main.swing/src/jsettlers/main/swing/foldertree/FileTreeCellRenderer.java
```diff
@@ -0,0 +1,109 @@
...
+ * @author Andreas Butti
+ *
+ */
+public class FileTreeCellRenderer extends DefaultTreeCellRenderer {
+	private static final long serialVersionUID = 1L;
+
+	/**
+	 * Filesystem view
+	 */
+	private final FileSystemView fileSystemView = FileSystemView.getFileSystemView();
```
andreas-eberle: Should this be static?

andreas-eberle:
@andreasb242: I'm just looking through the code and added some questions. As you cannot directly push to this branch, I suggest you answer the questions and I will adapt the changes (as long, as there are no major changes).

jsettlers.main.swing/src/jsettlers/main/swing/foldertree/FilesystemTreeNode.java
```diff
@@ -0,0 +1,277 @@
...
+	 * 
+	 * @param node
+	 *            Node
+	 */
+	public void add(FilesystemTreeNode node) {
+		children.add(node);
+	}
+
+	/**
+	 * Load the children to a tmp attribute, **this is the only one method which is called from another thread**
```
andreas-eberle: From which Thread is this method called? And on which Thread is the rest of the class running?

jsettlers.main.swing/src/jsettlers/main/swing/foldertree/SelectSettlerFolderDialog.java
andreasb242: For sure, the readme is also OK

jsettlers.main.swing/src/jsettlers/main/swing/foldertree/FileTreeCellRenderer.java
andreasb242: I know, for me the Constructor is the separator line between Attributes / Listener implementations and Methods, because of this I always add a constructor.

jsettlers.main.swing/src/jsettlers/main/swing/foldertree/FileTreeCellRenderer.java
andreasb242: Yes, the FileSystemView returns a static instance, so it can be static here also

jsettlers.main.swing/src/jsettlers/main/swing/foldertree/FilesystemTreeNode.java
andreasb242: This is only called for the root Folders, which are called in the UI Thead:
```java
		for (File f : File.listRoots()) {
			root.add(new FilesystemTreeNode(f));
		}
```
The asynchronous loading is working with this list
```java
	/**
	 * Tmp loaded child list
	 */
	private final List<FilesystemTreeNode> list = new ArrayList<>();
```
The list is applied to the main list in the method
```java
	/**
	 * Apply the loaded children
	 */
	public void applyLoadedChildren() {
		children = list;
	}
```
which is called from the UI Thread.
So everything should be synchronized.

andreas-eberle: Well, I just debugged into this and found out that this method was never called on another thread as 

```
	public void loadAsynchron(final FilesystemTreeNode node) {
		executor.submit(new Runnable() {

			@Override
			public void run() {
				node.loadChildren1();
                ....
```

was never called. Is this a bug, or does this show we don't need the async loading?

jsettlers.main.swing/src/jsettlers/main/swing/foldertree/FileTreeCellRenderer.java
andreas-eberle: OK, I understand your idea. However, I think it is better to have less 
code...

Am 28.02.2016 um 11:27 schrieb andreasb242:

jsettlers.main.swing/src/jsettlers/main/swing/foldertree/FilesystemTreeNode.java
andreasb242: This is a Bug, but my current system is running with an SSD, then the folder loading is usually that fast that you cannot recognize the difference.
To test add a small sleep in loadChildren1()

To fix the asynchron loading:
```java
	/**
	 * Add a node
	 * 
	 * @param node
	 *            Node
	 */
	public void add(FilesystemTreeNode node) {
		children.add(node);
		node.setParent(this);
	}
```
and in loadChildren1()
```java
			if (f.isDirectory()) {
				FilesystemTreeNode node = new FilesystemTreeNode(f);
				list.add(node);
				node.setParent(this);
```

Then it's working, you can check it with the sleep, the GUI is not freezing, even if it takes very long to load expanding folders.

jsettlers.main.swing/src/jsettlers/main/swing/foldertree/PathPanel.java
```diff
@@ -0,0 +1,153 @@
...
+			g.setStroke(new BasicStroke(2));
+			int w = getWidth();
+			int h = getHeight();
+			g.drawLine(0, 0, w - 1, h / 2);
+			g.drawLine(w - 1, h / 2, 0, h);
+		}
+	}
+
+	/**
+	 * Listener
```
andreas-eberle: Don't add javadoc just that there is a javadoc. In cases like this, where it is obvious, no javadoc is better than this one, because javadocs like this make the code less readable.

PR #311, opened 2016-03-15
Feature working donkeys

andreas-eberle:
Implementation of working donkeys and market.


jsettlers.logic/src/jsettlers/logic/buildings/Building.java
```diff
@@ -81,12 +84,12 @@
 	private static final EPriority[] SUPPORTED_PRIORITIES_FOR_CONSTRUCTION = new EPriority[] { EPriority.LOW, EPriority.HIGH, EPriority.STOPPED };
 	private static final EPriority[] SUPPORTED_PRIORITIES_FOR_NON_WORKERS = new EPriority[0];
 
-	private static final ConcurrentLinkedQueue<Building> allBuildings = new ConcurrentLinkedQueue<Building>();
+	private static final ConcurrentLinkedQueue<Building> ALL_BUILDINGS = new ConcurrentLinkedQueue<Building>();
```
homoroselaps: why is it Capitalized? Is it a Singleton?

andreas-eberle: In Java all constants are assigned upper case names (see http://www.oracle.com/technetwork/java/javase/documentation/codeconventions-135099.html#367).

michaelzangl: But this is not a constant. A constant is immutable.

This is a global reference to an object. The reference itself cannot be changed, but that is what final is there for. The object will be changing, so we should use lower case here.

andreas-eberle: Well, the coding convention states "The names of variables declared class constants".

michaelzangl: But it does not define the term "constant".

Definition by wikipedia: "In computer programming, a constant is an identifier with an associated value which cannot be altered by the program during normal execution"

That queue can be altered in a sense that the new value is not equal to the old value.

andreas-eberle: Well, the JLS states exactly: 
```
The names of constants in interface types should be, and final variables of class types may conventionally be, a sequence of one or more words, acronyms, or abbreviations, all uppercase, with components separated by underscore "_" characters.
```

=> We can choose the naming of `static final OBJECT_TYPE`. Therefore I think your idea to only use uppercase for really constant values is a good idea. I will change the naming.

@homoroselaps: Thnaks for bringing this up.

andreas-eberle:
@michaelzangl: There are some rendering issues with the market. Goods placed in the north west part of the market are hidden by it and donkeys and movables are sometimes also when walking in the north east. Can we change the rendering of it to prevent that?

JKatzwinkel:
Nice feature! Quick question, tho:
What are those market waypoints for? Are they for customizing the path a donkey is to take from one  market to another? Because I noticed that when the direct/bee-line way between two markets is blocked by e.g. water, donkeys seem to rather not to, and just drop their stuff immediately after taking off the market, only to turn around and pick up a new item repeatedly.
So I changed the path they'd supposedly go by putting way points, but then they would drop certain types of cargo at each of those way points and then walk back to their original market, never arriving at the destination market.

andreas-eberle:
Well, there is multiple things:
* You discovered a bug: It should not be possible to place the target or any waypoint on a blocked position (including buildings and lakes). I will fix that.
* I guess you selected another market as target location? The market center is blocked and therefore no one walk there... thats why the donkeys abort the walking and drop the stuff... (That's why you should not be able to place the target location there.)
* In Settlers III, the market can deliver goods anywhere. You usually do not deliver the goods to another market and you cannot connect two markets (that was the case in Settlers IV). 
* The waypoints are positions the donkeys pass on their route to the target location. This feature can be used to navigate them around enemies. That's important, because an enemy soldier can attack a donkey. If an enemy attacks a donkey, it will not be hurt (it cannot be killed) but it will drop all the goods at that location.

All in all, thanks for discovering the bug ;)

JKatzwinkel:
I am not sure if we perfectly understood each other nor if I am fully aware of how those things worked in Settlers III [I don't think I ever saw Settlers IV ever]... Let me try to address your questions...

  * I did not try to put a market destination point on a blocked tile. I'm sorry I arose the impression. I don't think I discovered that bug, but I'm happy if I was of help.
  * I did exclusively select other markets as market destination points. Because I thought I remembered this was how they worked from back in the time I played Settlers III. I was not aware that its the other way around and other markets were the one thing a market is not supposed to have a a destination...
  * Still, how would I be able to predict what type of cargo the donkeys will drop at what explicitly positioned waypoint? It appeared random to me.
  * The way you describe how to put the market waypoints and what for is exactly what I expected.

Another thing is that whenever I tried to use markets and donkeys, there was a lot of congestion going on between donkeys, donkeys and builders and stuff and I think even ordinary carriers. This can be caused by me having also merged the jam avoidance pull request already.
Because of two other possible bugs [I can't play the market test map because of out of bounds exception and I can't save a gameplay because of serialization error or whatever] it takes me a while to develop a single player game into a state where I can play around with markets, so I hope you take it from here.

homoroselaps:
If something goes wrong with the pathing, movables take damage and usually die (dunno exactly why). This could lead to cargo getting drop at waypoints.
________________________________

andreas-eberle:
@JKatzwinkel:
* Anyways, thanks for helping me find that missing part :D
* When you select the center of another market as target location, you actually selected a blocked location. 
* Markets can have any non-blocked position as target location. So you cannot select any buildings or so, as they are blocked. I will have to implement the check that prevents you from setting the target location to a blocked location.
* Donkeys should only drop the goods at the target location. The waypoints are only waypoints, meaning the donkeys have to pass them, but will not drop anything there.

The behavior you describe sounds very strange... Which branches did you exactly merge? Can you try to reproduce it with this (clean) branch? Please note that its in general a bad idea to combine various pull requests. Master is the only (kind of) stable branch.

JKatzwinkel:
I experienced this behaviour in a branch containing the swing GUI, the better jam avoidance and the donkeys pull requests. This immediately turned out to be a bad idea, however I was bored and managed to get it built. I will get back to you as soon as I tried using only donkey branch.

andreas-eberle:
@JKatzwinkel: I fixed the serialization problem you experienced.

JKatzwinkel:
@andreas-eberle Pulling now.

jsettlers.logic/src/jsettlers/logic/buildings/stack/multi/MultiRequestStack.java
```diff
@@ -0,0 +1,237 @@
...
+	}
+
+	@Override
+	public int getStackSize() {
+		return currentMaterialType == null ? 0 : grid.getStackSize(position, currentMaterialType);
+	}
+
+	@Override
+	public short getStillRequired() {
+		throw new UnsupportedOperationException();
```
homoroselaps: Should the message be "not implemented yet"?

jsettlers.logic/src/jsettlers/logic/movable/MovableStrategy.java
```diff
@@ -272,6 +275,10 @@ protected boolean isMoveToAble() {
 		return false;
 	}
 
+	protected boolean isAttackable() {
+		return movable.getMovableType().isMoveToAble();
```
homoroselaps: What exactly should isMoveToAble mean?

jsettlers.logic/src/jsettlers/logic/movable/strategies/DonkeyStrategy.java
```diff
@@ -0,0 +1,173 @@
...
+	@Override
+	protected boolean receiveHit() {
+		if (state == EDonkeyState.GOING_TO_TARGET) {
+			reset();
+			super.abortPath();
+		}
+		return false;
+	}
+
+	public static interface IDonkeyMarket extends ILocatable {
```
homoroselaps: Consider prepending it to the market or placing it in an extra file

jsettlers.tests/tools/jsettlers/logic/movable/MovableTestWindow.java
```diff
@@ -31,7 +31,7 @@
 
 public class MovableTestWindow {
 	private static final Player PLAYER_0 = new Player((byte) 0, new Team((byte) 0), (byte) 1);
-	private Movable movable;
+	private final Movable movable;
```
homoroselaps: should it be capitalized or not?

jsettlers.logic/src/jsettlers/logic/movable/MovableStrategy.java
michaelzangl: That the user can control it by doing a MOVE_TO-Command for this movable.

jsettlers.tests/tools/jsettlers/logic/movable/MovableTestWindow.java
andreas-eberle: This one is not a class constant => not capitalized.

JKatzwinkel:
Donkeys still drop material at waypoints occasionally. But I could not reproduce the circumstances under which this happens.

Also, maybe the "donkey always prefers his latest market" might be problematic, because if you build a new market, and all your donkeys work on a fully stocked market, it may take some time till a newly hatched donkey shows up and works your new market.

Apart from that, everything seems to work like a real charm and is a lot of fun!

jsettlers.logic/src/jsettlers/logic/buildings/stack/multi/MultiRequestStack.java
andreas-eberle: The problem is that this type of stack cannot support this method. In the case that no currentMaterialType is set, it is unclear what to return here. However, as this type of stack is never used during construction (which is the user of this method), it is not a problem that this method is not supported here.

Therefore I thought UnsupportedOperationException is more correct than "not implemented yet".

andreas-eberle:
@michaelzangl: The market UI does not show the button to change its priority. Can you add that button, so we can switch between the priorities? I already added the `getSupportedPriorities()` method to give you the priorities LOW, HIGH and STOPPED.

andreas-eberle:
@JKatzwinkel: Can you send me a replay with donkeys dropping stuff at waypoints? (You might need to rename it to .txt to be able to attach it to your post.) Please also add the revision you used. 

Regarding the last market strategy: I see your concerns... however, if you really want to share the donkeys, you can stop working of the market that has all the donkeys for a while (at least when @michaelzangl adds the button for it ;) ). When enough donkeys switched over to the other market, you can let it work again. I know this isn't perfect here, but I did not really come up with a better solution. 

The problem with randomly choosing the next market to go to, is that donkeys will often walk large distances between markets without carrying anything.

jsettlers.logic/src/jsettlers/logic/buildings/trading/MarketBuilding.java
```diff
@@ -0,0 +1,129 @@
...
+		@Override
+		public boolean hasNext() {
+			for (; i < waypoints.length && waypoints[i] == null; i++)
+				;
+			return i < waypoints.length;
+		}
+
+		@Override
+		public ShortPoint2D next() {
+			return waypoints[i++];
```
michaelzangl: Do a hasNext() here.

andreas-eberle: thx.

jsettlers.tools/testers/jsettlers/logic/movable/testmap/MovableTestsMap.java
```diff
@@ -1,450 +1,450 @@
-/*******************************************************************************
```
michaelzangl: Please change file to unix line endings

jsettlers.tools/testers/jsettlers/logic/movable/MovableTestWindow.java
```diff
@@ -1,105 +1,105 @@
-/*******************************************************************************
```
michaelzangl: Please change file to unix line endings

andreas-eberle: done

jsettlers.tools/testers/jsettlers/logic/movable/testmap/MovableTestsMap.java
andreas-eberle: done

michaelzangl:
This PR looks fine. Have not tested the new stuff on Android though (interface is still missing there.)

Issue #309, opened 2016-03-06
way finding algorithm results in jams


homoroselaps:
As the way finding mechanism chooses always the shortest route, this results in jams. The settles never choose a different one, even if they have to wait a long long time.
Builders in the way are a common problem, which causes the building process to slow down drastically.
![commonbuilderjam](https://cloud.githubusercontent.com/assets/7871498/13557335/d740aa42-e3ee-11e5-8bc5-5c9b5f57fa00.png)

homoroselaps:
My proposal: if a settler cannot walk to the next path for a long time, have a look for free patches next to it, choose one by random and calculate a new path.

andreas-eberle:
Yes, I know that problem very well... It can be divided into two problems: 

1. The situation with the builders should be locally fixable and normally, settlers should already walk around single other settlers standing in there way. I will have a look at that again.
 
2. The general issue with jams caused by many people using the same shortest path cannot be fixed that easily (at least I didn't find a solution with an acceptable performance). The key issue is that the A* algorithm often finds the same shortest paths for many settlers; although they might have differing start and target locations. This can for example be observed when sending a bunch of soldiers somewhere. I'm open for discussion and PullRequests regarding this problem ;)

andreas-eberle:
I was able to fix part 1 in #312.

andreas-eberle:
@homoroselaps: Why did you close this issue?

homoroselaps:
Because you fixed a part of it and there is already an issue out there which targets the main problem.

PR #317, opened 2016-03-26
Enhancement better jam avoidance

homoroselaps:
This fixes issue: #309
Implements a characteristic number for how much a movable is stuck in jam.
Every movable gets the possibility to request a new path if stuck. This time other stuck movables are treated as obstacles.
The frequency at which a movables is able to request a new path shrinks the more its stuck. (performance balanced in favor for desktop computer)


homoroselaps:
I found an performance + strategy issue if soldiers get stuck.

homoroselaps:
I am going to reimplement the logic in a cleaner way.

PR #326, opened 2016-03-28
Fix saving error by making private class serializable.

JKatzwinkel:
Private class `MarketBuilding$WaypointIterator` needs to be serializable, as it seems. With this modification, I can save games even if I got active markets.


andreas-eberle:
Thanks, that's right!

PR #328, opened 2016-03-30
clean up(rename/remove) the interfaces of Movable and MovableStrategy,

homoroselaps:


jsettlers.logic/src/jsettlers/logic/movable/Movable.java
```diff
@@ -204,8 +204,8 @@ public int timerEvent() {
 			break;
 		}
 
-		if (moveToRequest != null) {
-			if (strategy.isMoveToAble()) {
+		if (targetPosition != null) {
+			if (strategy.isAbleToMove()) {
```
michaelzangl: isMoveToAble is intended: is EActionType.MOVE_TO able ;-)

jsettlers.logic/src/jsettlers/logic/movable/Movable.java
```diff
@@ -78,7 +78,7 @@
 
 	private ShortPoint2D position;
 
-	private ShortPoint2D moveToRequest = null;
+	private ShortPoint2D targetPosition = null;
```
michaelzangl: moveToRequest is intended to be the position the user wants the movable to move to.

jsettlers.logic/src/jsettlers/logic/movable/Movable.java
```diff
@@ -548,7 +549,7 @@ final void drop(EMaterialType materialToDrop) {
 	 * @param sleepTime
 	 *            time to sleep in milliseconds
 	 */
-	final void wait(short sleepTime) {
+	final void sleep(short sleepTime) {
```
michaelzangl: good ;-)

jsettlers.logic/src/jsettlers/logic/movable/Movable.java
```diff
@@ -595,9 +596,9 @@ final boolean goToPos(ShortPoint2D targetPos) {
 	 *         false if the target position is generally blocked or a movable occupies that position.
 	 */
 	final boolean goInDirection(EDirection direction, boolean force) {
-		ShortPoint2D pos = direction.getNextHexPoint(position);
-		if (force || (grid.isValidPosition(this, pos) && grid.hasNoMovableAt(pos.x, pos.y))) {
-			initGoingSingleStep(pos);
+		ShortPoint2D targetPos = direction.getNextHexPoint(position);
```
michaelzangl: Good

jsettlers.logic/src/jsettlers/logic/movable/Movable.java
```diff
@@ -708,6 +692,13 @@ private void setState(EMovableState newState) {
 		this.state = newState;
 	}
 
+	private void setStrategy(MovableStrategy newStrategy) {
```
michaelzangl: It is always good to keep common things together. But why did you move only this method?

In my opinion, we can clear this class up even more:
- Separate static and non-static methods
- Group the general attribute getters, order by: id, movable type, attackable, player, state
- Group the whole strategy modification methods
- Group the selection methods (is/set selected)

jsettlers.logic/src/jsettlers/logic/movable/MovableStrategy.java
```diff
@@ -45,10 +46,12 @@
 public abstract class MovableStrategy implements Serializable {
 	private static final long serialVersionUID = 3135655342562634378L;
 
-	private final Movable movable;
+	protected final Movable movable;
+	protected final AbstractMovableGrid grid;
```
michaelzangl: What are you keeping this for?

jsettlers.logic/src/jsettlers/logic/movable/MovableStrategy.java
```diff
@@ -190,22 +176,18 @@ protected final void enableNothingToDoAction(boolean enable) {
...
-		movable.setSelected(selected);
-	}
-
 	protected final boolean fitsSearchType(ShortPoint2D pos, ESearchType searchType) {
-		return movable.getStrategyGrid().fitsSearchType(movable, pos, searchType);
-	}
-
-	protected final boolean isValidPosition(ShortPoint2D position) {
-		return movable.isValidPosition(position);
+		return movable.getGrid().fitsSearchType(movable, pos, searchType);
```
michaelzangl: Why not this.grid?

michaelzangl:
Thanks for your changes. Settlers is a grown project and we always welcome additions that make the code more readable.

I added notes on some name changes you made: Some names are there for only historical reasons, but some have a meaning (like isMoveToAble) that may not be seen easily. Feel free to add javadoc comments for those ;-)

You also moved some methods. This is a good thing to clean up the code, but can you please do a separate PR next time - best would be to not do both things at once to avoid conflicts. That way we can review it more easily.

jsettlers.logic/src/jsettlers/logic/movable/Movable.java
homoroselaps: so targetPosition ok, isnt it? As its not some sort of  "Request Object" I think this might be better.

jsettlers.logic/src/jsettlers/logic/movable/MovableStrategy.java
homoroselaps: thx

jsettlers.logic/src/jsettlers/logic/movable/MovableStrategy.java
homoroselaps: to keep the code in the other strategies more lean.

homoroselaps:
@michaelzangl I see what you mean. But at some point working with too many PR branches its just not fun.
May you give me advice how to handle the failed tests? I didnt change the AI logic and all checks fail on travis.

jsettlers.logic/src/jsettlers/logic/movable/Movable.java
michaelzangl: We can also name it requestedTargetPosition.

It is a request by the user - not to be confused with the target the pathfinder uses (e.g. soldier attacking, ...).

jsettlers.logic/src/jsettlers/logic/movable/MovableStrategy.java
michaelzangl: I don't see a big difference between getGrid().... and grid....
I think we should not add this field simply to make the code a few characters shorter.

michaelzangl:
I know that problem. I try to separate that in time: Add an issue that the reordering needs to be done as soon as your first PR is accepted. You can suggest an order there. That way you get feedback and we can reorder the methods afterwards using a new PR.

For the failed test: As soon as you rename fields, the saved files change since they contain serialized versions of those objects. This breaks the ability to load old files and to play over the network with older versions. This is why we try to avoid renaming fields and classes if it is not necessary.

jsettlers.logic/src/jsettlers/logic/movable/Movable.java
```diff
@@ -60,9 +60,30 @@
 	private static final ConcurrentLinkedQueue<Movable> allMovables = new ConcurrentLinkedQueue<Movable>();
```
michaelzangl: Please do not do that ;-)

Only reorder or re-format if there is a real significance of doing it. This will only lead to merge conflicts with other branches, especially the donkey branch.

jsettlers.logic/src/jsettlers/logic/movable/Movable.java
andreas-eberle: No, please don't reorder all the methods...
1. his makes it impossible to somehow review this PR
2. most of the internal methods are grouped together because they belong together and are used by each other. Therefore this makes reading, understanding and working in the code much easier as when simply grouping the methods. When reordering something, be aware that the order you see is mostly intended.

jsettlers.logic/src/jsettlers/logic/movable/MovableStrategy.java
homoroselaps: As it is used so often, its cleaner to keep it and we save some function calls ^^.

michaelzangl: I don't think that is cleaner to have redundant data.

Don't worry about that method call. This is Java, not C++. We have a compiler to worry about that ;-).

homoroselaps:
@andreas-eberle What are the next steps to get this PR accepted?

andreas-eberle:
Well, there is two things: 
1. I need to finish reviewing the new swing ui PR and merge it.
2. I wasn't able to take a deeper look at this PR yet. However, I saw that some tests are failing. Of course, these need to be fixed. 

Regarding the tests:
* The AutoReplayIT tests have fixed replay files that are replayed by the test and the resulting game is saved. The savegame is then compared with the referene file commited into the repository. As Michael already pointed out, these tests break if you rename/move classes or rename/add/remove fields. Of course this is a legitimate step during refactorings. Therefore you can regenerate the reference files (and commit them) by running the main method of the AutoReplayIT class. However, before you commit those reference files, be really sure you didn't break the game logic (e.g. by playing the game for some game hours).
* The AiDiificultiesIT is different. Here, no savegames are stored and compared. These tests only check if the stronger AIs are able to beat the weaker ones and if they produce a certain amount of gold in a given time (which is kind of a economy benchmark). Therefore these tests must never fail due to refactorings. Therefore this test luckily shows you that you somehow changed the game logic. This does not neccisarly mean that you broke it, but in a refactoring PR, no logic changes should be included. 

All in all, please have a look at the failing tests and find out what changed the game logic (and fix that change).

I hope to be able to review the PRs more in deep during the next week(s). Unfortunately I have an exam comming up... So I won't be able to work as much on the project as I want to...

jsettlers.logic/src/jsettlers/logic/movable/Movable.java
```diff
@@ -606,23 +607,10 @@ final boolean goInDirection(EDirection direction, boolean force) {
 	}
 
 	/**
-	 * Forces the movable to go a step in the given direction (if it is not blocked).
-	 * 
-	 * @param direction
-	 *            direction to go
-	 */
-	final void forceGoInDirection(EDirection direction) {
```
andreas-eberle: This method does something different than goInDirection(direction,true). It's true, the naming is bullshit here, as both are called `force`. However, I think that both functionalities are needed... but the names should definitely be changed. I'm quite sure that this is a game logic change.

andreas-eberle:
I implemented the refactorings of this PR in #355. Many thanks to @homoroselaps for bringing this up. I will close this PR in favor of #355.

@homoroselaps: For future refactoring PRs, take a look at how I added the changes in #355. There, I strictly separated simple refactorings (not chaning the savegame) from changes that change the savegame but should not change the game logic. With this, it is possible to easily review the PR and see that no logic changes have been introduced, although the AutoReplayIT references had to be adjusted.

PR #329, opened 2016-03-30
clean up (Rename/Remove) of Building

homoroselaps:
and IConstructableBuilding/IWorkerRequestBuilding

### DIFF
--- a/jsettlers.common/src/jsettlers/common/buildings/jobs/IBuildingJob.java
+++ b/jsettlers.common/src/jsettlers/common/buildings/jobs/IBuildingJob.java
@@ -17,6 +17,7 @@ package jsettlers.common.buildings.jobs;
 import jsettlers.common.material.EMaterialType;
 import jsettlers.common.material.ESearchType;
 import jsettlers.common.movable.EDirection;
+import jsettlers.common.position.ShortPoint2D;
 
 /**
  * This is a building job, a job that can be done by a worker.
@@ -68,6 +69,14 @@ public interface IBuildingJob {
 	 * @return The y distance.
 	 */
 	short getDy();
+
+	/**
+	 * calculates the point on the grid from the given start point.
+	 * 
+	 * @param start
+	 * @return
+	 */
+	public ShortPoint2D calculatePoint(ShortPoint2D start);
 
 	/**
 	 * Gets the direction associated with this job.
--- a/jsettlers.logic/src/jsettlers/logic/buildings/Building.java
+++ b/jsettlers.logic/src/jsettlers/logic/buildings/Building.java
@@ -213,7 +213,8 @@ public abstract class Building extends AbstractHexMapObject implements IConstruc
 		return couldBePlaced;
 	}
 
-	protected abstract void placedAtEvent(ShortPoint2D pos);
+	protected void placedAtEvent(ShortPoint2D pos) {
+	};
 
 	/**
 	 * Used to set or clear the small red flag atop a building to indicate it is occupied.
--- a/jsettlers.logic/src/jsettlers/logic/buildings/Building.java
+++ b/jsettlers.logic/src/jsettlers/logic/buildings/Building.java
@@ -62,14 +62,8 @@ import jsettlers.logic.timer.RescheduleTimer;
 
 public abstract class Building extends AbstractHexMapObject implements IConstructableBuilding, IPlayerable, IBuilding, IScheduledTimerable,
 		IDebugable, IDiggerRequester, IViewDistancable {
-	private static final long serialVersionUID = 4379555028512391595L;
 
-	private static final byte STATE_CREATED = 0;
-	private static final byte STATE_IN_FLATTERNING = 1;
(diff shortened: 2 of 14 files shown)


jsettlers.logic/src/jsettlers/logic/buildings/Building.java
```diff
@@ -88,7 +82,7 @@
 	private ShortPoint2D pos;
 	private IBuildingsGrid grid;
 	private Player player;
-	private byte state = STATE_CREATED;
+	private EBuildingState state = EBuildingState.CREATED;
```
michaelzangl: Nice, thanks

jsettlers.logic/src/jsettlers/logic/buildings/Building.java
```diff
@@ -715,4 +704,13 @@ public short getPartitionId() {
 	public boolean cannotWork() {
 		return false;
 	}
+
+	public enum EBuildingState {
```
michaelzangl: Can you please add a comment here that the order is important and should not be changed - otherwise savegames will break.

jsettlers.logic/src/jsettlers/logic/buildings/Building.java
```diff
@@ -201,42 +198,39 @@ protected void placeAdditionalMapObjects(IBuildingsGrid grid, ShortPoint2D pos,
...
+				showFlag(true);
 			}
 
-			positionedEvent(pos);
+			placedAtEvent(pos);
 		}
 		return couldBePlaced;
 	}
 
+	protected abstract void placedAtEvent(ShortPoint2D pos);
```
michaelzangl: Why not remove the abstract modifier?

jsettlers.logic/src/jsettlers/logic/movable/strategies/BuildingWorkerStrategy.java
```diff
@@ -393,7 +394,7 @@ private void jobFailed() {
 	}
 
 	private ShortPoint2D getCurrentJobPos() {
-		return building.calculateRealPoint(currentJob.getDx(), currentJob.getDy());
+		return new RelativePoint(currentJob.getDx(), currentJob.getDy()).calculatePoint(building.getPos());
```
michaelzangl: The JIT should be able to optimize this in such simple cases, so this is OK for PC. The Android JIT seems to have more problems optimizing this and removing the object allocation.

Changes like this are always dangerous in that they might introduce a massive creation of short-lived objects. We should not worry about this too much but this was the case why we use calculateRealPoint.

But you could make the job class extend RelativePoint, that way you save the extra object allocation and code gets cleaner.

michaelzangl:
I really like this one. Thanks.

jsettlers.logic/src/jsettlers/logic/buildings/Building.java
homoroselaps: I dont understand why you are so careful not to break replays and savegames?! I usually play one scenario until I win or lose - and would never in between checkout the newest version from github, compile it and keep playing... I thought thats why there are official builds?!

michaelzangl: We use those a lot for testing.

Some of the savegames and replay files are in the repository, I have some more locally. This makes developing the UI faster - simply load the savegame and see if all images look the way they should. No need to play through a whole game.

This is why every update that breaks savegames also breaks local development and our automated tests. Therefore we avoid it if it is not necessary.

jsettlers.logic/src/jsettlers/logic/movable/strategies/BuildingWorkerStrategy.java
homoroselaps: - no comment -
*nice idea.

jsettlers.logic/src/jsettlers/logic/buildings/Building.java
```diff
@@ -213,7 +213,8 @@ private boolean placeAt(IBuildingsGrid grid, ShortPoint2D pos) {
 		return couldBePlaced;
 	}
 
-	protected abstract void placedAtEvent(ShortPoint2D pos);
+	protected void placedAtEvent(ShortPoint2D pos) {
```
michaelzangl: Can you please remove the empty method stubs in subclasses?

jsettlers.common/src/jsettlers/common/buildings/jobs/IBuildingJob.java
```diff
@@ -70,6 +71,14 @@
 	short getDy();
 
 	/**
+	 * calculates the point on the grid from the given start point.
+	 * 
+	 * @param start
+	 * @return
+	 */
+	public ShortPoint2D calculatePoint(ShortPoint2D start);
```
michaelzangl: Nice, thanks.

jsettlers.logic/src/jsettlers/logic/buildings/Building.java
homoroselaps: thx - I understand that.

Issue #280, opened 2016-01-11
Do not print the same message multiple times.


DerKrawallkeks:
Hey:)
I don't think it's necessary to have a notification for each hit that a soldier takes. Doesn't sound good and spams the notification section. By the way..please make notifications disappear after a few seconds if possible:)

Zervox:
it does the same on PC, besides this is under development, it's the same thing with woodworker can't find work etc it gets spammed, I think this is done atm for debugging.

DerKrawallkeks:
Okay, thanks

JKatzwinkel:
Avoiding rapidly repeated messages from buildings is not a problem because buildings don't move. But attack messages don't come from stationary positions, and this necessitates heuristics for determining whether an incoming message reports a known incident or a different attack. I feel like I achieved satisfactory results by ignoring all attack messages that originate from within a [*arbitrarily chosen*] range of 50 tiles around where the latest reported attack took place.

Removing messages older than 60 seconds makes sure that even redundant attack messages are shown at least once a minute, in case someone manages to miss it the first time.

andreas-eberle:
Fixed with #331.

PR #331, opened 2016-03-31
Enhancement less messages

JKatzwinkel:
These changes address issue #280. Flooding of the message queue by repetition of the same or redundant messages is significantly reduced. This is done by imposing the following requirements on messages before they are shown:

  * message is not older than 90 seconds.
  * message does not come from the same sender and location nor does it have the same type and content as any of the messages currently shown.
  * a reported attack didn't occur within a small radius of any attacks by the same player reported by the currently shown messages.
  * same goes for resource discoveries, but those don't seem to be reported atm.

To indicate how much time has passed since a message has been reported, its text is now blitted with slowly decreasing opacity, after blinking for 1.5 seconds to get attention.
Because this did not work with the black rectangles they used to be put on for better visibility, those have been replaced by shadow text. This means there are a lot of text drawing and color setting operations going on compared to before. I don't know if this is a performance problem.

I also made message text bigger and moved it a bit to the right and increased the number of concurrently shown messages.

Audio notification of an attack should also be played only if the message fits the above criteria. But I have no pulseaudio, so not sure if this works like expected.


jsettlers.graphics/src/jsettlers/graphics/map/MapContent.java
```diff
@@ -291,27 +301,28 @@ private void drawMessages(GLDrawContext gl) {
...
+				drawer.setColor(1, 1, 1, a);
 				drawer.drawString(x, y, m.getMessage());
 
 				messageIndex++;
-				if (messageIndex >= MAX_MESSAGES) {
+				if (messageIndex >= Messenger.MAX_MESSAGES) {
 					break;
 				}
 			}
+			drawer.setColor(1, 1, 1, 1);
```
michaelzangl: You don't need to set this. Every component should set the font color before drawing.

jsettlers.graphics/src/jsettlers/graphics/messages/Messenger.java
```diff
@@ -37,10 +49,42 @@
...
+			messages.addFirst(message);
+			if (messages.size() > MAX_MESSAGES)
+				messages.removeLast();
+			return true;
+		}
+		return false;
+	}
+
+	public void removeOld() {
+		for (; (messages.size() > 0)
```
michaelzangl: Please convert this to a while loop

jsettlers.graphics/src/jsettlers/graphics/messages/Messenger.java
```diff
@@ -37,10 +49,42 @@
...
+				&& (messages.getLast().getAge()>MESSAGE_TTL);
+				messages.removeLast());
+	}
+
+	boolean isNews(Message msg) {
+		if (messages.isEmpty()) {
+			return true;
+		}
+		for (Message m : messages) {
+			if ((m.getSender() == msg.getSender())
```
michaelzangl: Can you move this to the message class?

Something like Message#isLike(otherMessage)

This would make the code cleaner.

jsettlers.graphics/src/jsettlers/graphics/messages/SimpleMessage.java
```diff
@@ -46,7 +48,7 @@ public EMessageType getType() {
 	@Override
 	public long getAge() {
 		// TODO: implement a message aging process.
```
michaelzangl: Remove TODO (done)

jsettlers.graphics/src/jsettlers/graphics/messages/SimpleMessage.java
```diff
@@ -79,4 +81,16 @@ public static Message cannotFindWork(IBuilding building) {
 		String message = Labels.getString("cannot_find_work_" + building.getBuildingType());
 		return new SimpleMessage(EMessageType.NOTHING_FOUND_IN_SEARCH_AREA, message, (byte) -1, building.getPos());
 	}
+
+	@Override
+	public boolean equals(Object obj) {
```
michaelzangl: Always override hashCode when overriding equals.

michaelzangl:
Thanks for the patch. I added some minor notes on your patch.

In general it looks fine. Performance of the text draws should not be a big issue since you always draw the same string. Both Android and JOGL cache the text on a OpenGL texture. That way, rendering the same text multiple times is no problem.

jsettlers.graphics/src/jsettlers/graphics/map/MapContent.java
JKatzwinkel: Apparently, not all components do . When I left this out, menu panel text became invisible.

michaelzangl: OK, this is a bug then...

At least add a TODO that this can be removed once that bug is solved.

jsettlers.graphics/src/jsettlers/graphics/messages/Messenger.java
JKatzwinkel: I was thinking about it (*as well as I considered putting the comparison code into `equals` and just calls `contains(message)`*). Will do that.

JKatzwinkel:
Thanks for your useful remarks. While I'm at it, should I change message age determination so that it uses `MatchConstants.clock().getTime()`?

michaelzangl:
Adding this is not the real problem.

This depends on whether you want to use real or game time to hide the messages.

Should they stay there even if the game pauses? Should they hide slower if the game runs slower? Imho, we can simply take real time.

I am more concerned about calling System.currentTimeMillis() so often. We will see if this is a problem when profiling, so don't worry about it now.

jsettlers.graphics/src/jsettlers/graphics/messages/Messenger.java
michaelzangl: This is very bad practice, since your equals is not transitive then.

jsettlers.graphics/src/jsettlers/graphics/map/MapContent.java
JKatzwinkel: What I said there isn't accurate. Must have been related to eclipse debug mode synching and doesn't occur anymore.

JKatzwinkel:
The reason why I asked is that when people pause a game, they probably do it because they are preoccupied with something else for a while. So when they come back from whatever that was, they won't see any messages anymore if we use system time.
I think system time also can cause problem if you use it to measure intervals (*daylight saving time etc.*).

Calling `currentTimeMillis` this frequently feels uncomfortable. Even on recent JVMs, its cost probably varies depending on OS and hardware. On the other hand, native implementations seems to have improved by a lot over the last years, and today CPU's have things like TSC and stuff. I don't know...

I can't build with `jsettlers.logic.constants` in the build path, so I couldn't try game clock yet.

michaelzangl:
Normally, we do not expose that directly.

We expose it indirectly: The messages already have a getAge() method. You now need a new interface (IClock) that has a getTime() method. This clock is then passed on to the message.

That way we can easily interchange different times.

The problem I see is when the game runs fast: When pressing F12, you skip all messages as well. If the game runs fast, you only see the messages very shortly and get spammed by geologist messages.

This is why I would suggest adding an other `float getNormalizedTime()` method to IClock. This method returns the current time the game was running (not paused, not in F12) normalized to normal speed in seconds.

I would implement it as follows:
- Have a way to access the time scale (e.g. 20 ticks per second)
- Set float normalizedTimeOffset = 0 at game start
- Set int normalizedTimeReferenceTick = 0 at game start
- On every speed change (including start of fast forward), set:
  - normalizedTimeOffset = getNormalizedTime()
  - normalizedTimeReferenceTick = getTime()
- After fast forward:
  - normalizedTimeReferenceTick = getTime()
- getNormalizedTime():
  - return inFastForward ? normalizedTimeOffset : normalizedTimeOffset + (getTime() - normalizedTimeReferenceTick) * currentSpeed

andreas-eberle:
I would not make it that complicated... and I would not expose the game time to the UI that way.

As @michaelzangl said: The IMessage

andreas-eberle:
I would not make it that complicated:

1. If you connect the message "age" to the game time, you will at least need to handle fast forward and faster game speed in single player. This makes it complicated (as you can see in @michaelzangl's message).
2. I would use the real time for message displaying duration. From reading your discussion, I only see two issues with that: 
   1. When the game is paused for a while, the messages would dissapear. Well, you simply need to get the Info if the game is currently paused. If it is, just don't "increase the alive timer" of the messages.
   2. It seams you're calling `currentTimeMillis` a lot atm. My proposal (without looking at the code) is to not accurately calculating the age of the messages by subtracting the start time from the current time. Instead, calculate the time since the last frame (or calculating of the message ages) once and add that value to an age counter of every message. With this, you still get a good enough age (we don't need exact values here), have a good performance and do not need a coupling with the game timer.

What do you think?

JKatzwinkel:
@andreas-eberle what would be the easiest way for an `Messenger` object to obtain the information whether the game is currently paused?

andreas-eberle:
I'm writing you a small PR to provide you the stuff you need.

andreas-eberle:
@JKatzwinkel: You got a PR. I removed the obsolete IStatiscable interface (which was only providing the game time) and replaced it with an IGameTimeProvider interface. I also gave it to the Messenger. So now you have the game time and the pausing state.

JKatzwinkel:
Nice! Thank you!

andreas-eberle:
@JKatzwinkel: Leave me a note, when you think you're ready to be reviewed.

JKatzwinkel:
@andreas-eberle I have been using the methods provided by your interface. However, determining message age based on the game clock causes their visual representation to become somewhat strange at accelerated/slowed-down game speed. I've switched back to use `currentTimeMillis` again (*calling it significantly less frequently*) and I find it way more appealing that messages fade out and get repeated at a constant pace.

@michaelzangl because I limit the number of calls to one per cycle, I feel ok still using system time. So I haven't implemented your proposal so far. My current implementation should make sure there is neither message spamming nor skipping while at the same time, messages won't easily slip your attention. Thanks for the idea, tho!

I started a new branch containing a keybinding that allows you to view the locations that come with the current messages. Its very useful obviously for finding that certain woodcutter who ran out of wood, obviously. Should I merge that into this or do you consider it a separate feature?

andreas-eberle:
@JKatzwinkel: I also think it's better to use `currentTimeMillis`. And calling it once per frame is fine. 
Regarding the other feature: Please create a new branch and Pull Request for it. It makes things easier to review.

JKatzwinkel:
@andreas-eberle @michaelzangl Test case `testCloseOneSide` failed at Travis', but I swear it did pass on my machine.
As to this PR, I'm not working on any further modifications right now, so what do you think?

michaelzangl:
It looks fine to me. We can always improve it afterwards if we see the need for it.

I just punched travis to run that test again.

andreas-eberle:
@JKatzwinkel: Some network unit tests are a bit fragile... sometimes they need to be retriggered. Just leave us a note if this happens.

PR #334, opened 2016-04-02
Remove config.prp from git

michaelzangl:
The config.prp file is always a source of merge conflicts. It is not good to have local settings in a file that is under version control.

I removed config.prp from version control.

There is a new template file that contains the default values. It contains a note on where to copy it if a user wants to edit that file manually.


michaelzangl:
This PR can be ignored for now. If we switch to gradle, I would update the whole config thing (resources is configured automatically and put the gfx/snd paths to <java.home>/.jsettlers)

michaelzangl:
Closing this since the move to gradle changed everything again and removed that config.prp template.

PR #341, opened 2016-04-05
Improve image preload cache.

michaelzangl:
This improves the way image preloading uses the image cache.

It also reduces memory consumption of graphics while the game is running since textures can be re-loaded from disk on demand (this should not happen in normal gameplay but e.g. after suspending the android app)


michaelzangl:
Github seems to be slow on catching up. This should contain 3 commits.

andreas-eberle:
@michaelzangl: Do you want to update this branch so it can be merged, or wait until after the gradle merge?

michaelzangl:
I'd like to wait until after the gradle branch is merged.

andreas-eberle:
@michaelzangl: Can you update this branch?

PR #337, opened 2016-04-04
Android building menu

michaelzangl:
Fixes the images and building counts for the build menu.

PR #342, opened 2016-04-06
Android remove image generator

michaelzangl:
Removes the image generator - we do not need it any more.


michaelzangl:
Requires #337

michaelzangl:
Same as #337

PR #349, opened 2016-04-08
Sort saved games antichronologically.

### LINKED ISSUE #343: Sort order of saved games.
With the new GUI (*`new_swing_gui` branch*), the list of saved maps is no longer in antichronological order. Instead, no sorting seems to be being applied at all, map order is the same as in `ls -f`.

JKatzwinkel:
This fixes #343.


andreas-eberle:
@JKatzwinkel: Thanks for your PR. However, while I had a look at it, I found more things needed to be fixed / done a bit different in that code part. That's why I opened #352 which replaces this PR.

JKatzwinkel:
@andreas-eberle I was hoping to learn what your solution would look like. Thanks for the insight and for combining with #351.

PR #350, opened 2016-04-08
Fix line endings - .classpath

michaelzangl:
Fix line endings in .classpath files.

Fix .gitattributes


.gitattributes
```diff
@@ -1,5 +1,5 @@
 # Set the default behavior, in case people don't have core.autocrlf set.
-* text=auto
+* text eol=lf
```
andreas-eberle: I think this is not a good idea. From my understanding, `auto` ensures that checked in files have LF line endings (therefore the repo has LF line endings). But on my local machine, I have the line ending that is appropriate for the system. Git will convert the line endings when commiting. 

Setting the `eol=lf`means that Windows developers will also have lf line endings on their local machine, which can have disadvantages. 

Therefore I suggest staying with `auto`.

.gitattributes
```diff
@@ -1,5 +1,5 @@
...
 
 # Declare files that will always have CRLF line endings on checkout.
-*.prp text eol=crlf
\ No newline at end of file
+*.prp text eol=crlf
```
michaelzangl: This enforces line endings to be crlf on all systems.

.gitattributes
michaelzangl: The problem I have with auto is the packing of resources: We want resource files to have \n line endings. If we do a release from a Windows machine, we might get problems.

I don't trust the git auto setting.

If you want to, we can set this explicitly for e.g. java files:
```
.java text
```

I believe that developers use a good text editor that can handle this and not the Windows default editor. So I think that we should stick to \n to prevent any further line ending problems.

.gitattributes
andreas-eberle: Only for the .prp files. And that is to ensure that if you build a release on any platform, the config.prp files must be readable for any user. This can only be ensured by enforcing `crlf` for the config.prp files.

.gitattributes
andreas-eberle: ok for me.

PR #327, opened 2016-03-30
Small Refactoring - IMapObject: Tree drawing

homoroselaps:
move constants away from an interface to the only point there are used

PR #355, opened 2016-04-10
Some refactorings to cleanup code suggested by @homoroselaps 

andreas-eberle:
as suggested by @homoroselaps in #329 and #327.

@homoroselaps: I wasn't able to merge your changes with some other changes I made in the Building class that changed the way Building objects were created. Therefore I did your changes from new on the current master. I will close your outdated PR in favor of this one. Nethertheless, many thanks for contributing these refactorings!


andreas-eberle:
I also integrated #327 as it only needed to be updated to the current master. I will therefore close #327.

andreas-eberle:
I think I got all the refactorings of @homoroselaps. @michaelzangl: Can you have a look at the changes?

PR #372, opened 2016-04-25
Extract autoreplay reference generator and fix some gradle stuff

andreas-eberle:
This PR fixes some gradle stuff
* correct clean in jsettlers.common
* removed depenency of jsettlers.logic:test on jsettlers.common:test
* made jsettlers.common:test actually execute the jsettlers.common tests

and it makes it possible to generate auto replay references again. This can be done by running the main method of jsettlers.tools/jsettlers.integration.replay.RegenerateAutoReplayITReferences class.


andreas-eberle:

Issue #219, opened 2015-10-27
Cannot build on top of mined stone


abledbody:
Stone that has been completely mined out cannot have buildings place on top of them, like in the original game.

andreas-eberle:
Can you add a screenshot showing this behavour? I wasn't able to repdoduce it.

abledbody:
@andreas-eberle ![Screenshot](http://puu.sh/l0Bg0/331b529c84.png)
It's possible this is either exclusive to map editor placed stone, or stonecutters can mine stone rubble.

michaelzangl:
I am more concerned that you can build on the tower in that screenshot.

andreas-eberle:
@abledbody: Yes, it seems to be an issue with the stones placed by the map editor. I will check that.
@michaelzangl: What do you mean? You can't build a tower on top of the other tower, but behind it...

michaelzangl:
Yes, the right tower. My bad, I'll blame the one that implemented the preview function.

andreas-eberle:
Yes, this was fixed. Thanks for closing the issue.

Issue #293, opened 2016-01-19
Stone is not harvested (@ 2Hours.map) 


tomsoftware:
In the original map [Siedler3\Map\SINGLE\2Hours.map] one stone can't be harvesded (see image)
[tested with different stonecutter-huts]

=> the hut calls: "This Stonecutter can't find no more stone"

![stome](https://cloud.githubusercontent.com/assets/3290888/12417124/29a5a148-bea6-11e5-94f1-6e78f828f2fa.jpg)

PR #378, opened 2016-04-26
Fixes uncuttable stone (#293)

andreas-eberle:
This PR fixes

* never dissapearing and blocking fully cut off stone which appeared if the loaded map contained fully cut off stone.
* corrected positioning of map object used for fully cut off stone
* implemented rendering for that map object type
* fixed removal of map objects when the ground is flattened for a building.


andreas-eberle:
Also fixes #219.

andreas-eberle:
@michaelzangl: Can you take a look at the changes?

jsettlers.common/src/main/java/jsettlers/common/mapobject/EMapObjectType.java
```diff
@@ -146,7 +150,21 @@
 	public static final EMapObjectType[] VALUES = EMapObjectType.values();
 	public final byte ordinal;
 
-	private EMapObjectType() {
+	public static final Set<EMapObjectType> TO_BE_REMOVED_WHEN_FLATTENED = new HashSet<>(Arrays.asList(
```
michaelzangl: Use EnumSet.of(...)

jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/MapObjectDrawer.java
```diff
@@ -338,10 +342,6 @@ private void drawObject(int x, int y, IMapObject object, float color) {
 			drawByProgress(x, y, 13, 42, progress, color);
 			break;
 
-		case WINE:
```
michaelzangl: Where did this go?

michaelzangl:
The rest looks fine to me.

jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/MapObjectDrawer.java
andreas-eberle: It was used a long time ago. But we then introduced WINE_GROWING, WINE_HARVESTABLE, WINE_DEAD (like it is also done for CROP_*)

jsettlers.common/src/main/java/jsettlers/common/mapobject/EMapObjectType.java
andreas-eberle: thx.

PR #380, opened 2016-04-26
Fixed rendering crash when preview building was activated and mouse left the map in direction north (fixes #379).

### LINKED ISSUE #379: Rendering crash when preview building is outside the map
When you select a building to be build, so you can see the preview building, the rendering will crash when you leave the map with the cursor to the north. Rendering crash means only the background (not even the menu) is drawn.

andreas-eberle:


jsettlers.graphics/src/main/java/jsettlers/graphics/map/MapContent.java
```diff
@@ -471,10 +470,12 @@ private void drawMain(FloatRectangle screen) {
 
 		if (placementBuilding != null) {
 			ShortPoint2D underMouse = this.context.getPositionOnScreen((float) mousePosition.getX(), (float) mousePosition.getY());
-			IMapObject mapObject = context.getMap().getMapObjectsAt(underMouse.x, underMouse.y);
+			if (0 <= underMouse.x && underMouse.x < width && 0 <= underMouse.y && underMouse.y < height) {
```
michaelzangl: Don't we have access to a containsPoint() here?

andreas-eberle: contains point checks if a screen position is on the UI (regarding the javadoc and judging by its usage).

PR #411, opened 2016-05-05
Select soldiers branch wise on double click.

andreas-eberle:


andreas-eberle:

jsettlers.logic/src/main/java/jsettlers/input/GuiInterface.java
```diff
@@ -572,18 +573,31 @@ private IGuiMovable getSelectableMovable(short x, short y) {
...
 			setSelection(new SelectionSet());
 			return;
 		}
 
-		final List<ISelectable> selected = new LinkedList<ISelectable>();
-	
+		EMovableType selectedType = selectedMovable.getMovableType();
+
+		Set<EMovableType> selectableTypes;
+		if (selectedType.isSwordsman()) {
```
michaelzangl: We do not need that case work here. Have a look at the android-selection branch. I improved the ESoldierType. You could add a getSoldierType() method to EMovableType and then compare the soldier type.

andreas-eberle: But what would I do with the normal movables? They cannot have a soldier type, so I would need a special case handling for them...

michaelzangl: Yes, you would and you can still leave the else part. This should just replace the isPikeman()..

You can then get the movable set from the ESoldierType.

andreas-eberle: I think we should stick with this implementation for now. When we extend the implementations of ESoldierType and ESoldierClass, we can use those functions here as well.

Issue #381, opened 2016-04-26
Implement logic for setting the amount of soldiers in a tower.


andreas-eberle:
Buttons should already work in SwingUI.

andreas-eberle:
Implemented with PR #418.

PR #418, opened 2016-05-06
Allow setting the soldiers of a tower

### LINKED ISSUE #415: Towers should search for all required soldier types simulatneously

andreas-eberle:
fixes #415, #381


andreas-eberle:
@codingberlin: One of the AI tests is failing. I guess this is because the towers are not completely filled with soldiers by default. Can you update the AI to set the towers to be fully occupied?

andreas-eberle:

jsettlers.common/src/main/java/jsettlers/common/utils/collections/map/OrderedMap.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Created by Andreas Eberle on 06.05.2016.
+ */
+public class OrderedMap<K, V> implements Serializable {
```
michaelzangl: What is the difference to LinkedHashSet?

michaelzangl:
The rest seems fine to me.

jsettlers.common/src/main/java/jsettlers/common/utils/collections/map/OrderedMap.java
andreas-eberle: The problem with LinkedHashMap is that it doesn't serialize consistently. If you serialize, deserialize and serialize a LinkedHashMap again, the bytes of the two serializations can be different. However, we need them to be the same to be able to see if savegames differ due to actual changes.

PR #424, opened 2016-05-06
Fixed request stacks and amount of requested materials for big temple

### LINKED ISSUE #423: Large temple shows plank twice under the needed materials.
See screenshot.

unbenannt

andreas-eberle:
fixes #423


jsettlers.common/src/main/resources/jsettlers/common/buildings/big_temple.xml
```diff
@@ -148,9 +148,8 @@
 
 	<door dx="0" dy="2" />
 
-	<constructionStack dx="5" dy="6" material="STONE" buildrequired="8" />
+	<constructionStack dx="5" dy="6" material="STONE" buildrequired="14" />
```
michaelzangl: Can't we leave this to be two stacks?

This is something that annoyed me in the original game - you always need to wait until the rest of the materials are brought to the construction site.

andreas-eberle: I don't think this is a good idea. This would change the balance and feeling of the game quite a bit. Furthermore, you usually don't have so much building materials available that you have to "store" them at construction sites. If you do, you should probably build more...

PR #427, opened 2016-05-08
AI Enhancements

codingberlin:
- AI uses geologists
- EASY, HARD, VERY_HARD does rush defence (tested with lv1 and lv2 12 weapons rush)
- reduce number of winegrowers to 3 which is enough for fast lv3
- optimized building lists to fasten the AIs
- introduce AiMapInformation to get the maximal amount of buildings for the map depending on the resources and size of the map
- replaced MiddleEconomyMinister and WinnerEconomyMinister by BuildingListEconomyMinister which builds the best possible long term economy depending on the AiMapInformation. For EASY and HARD it is downsized by a configurable factor which reduces the maximum number of weapon smiths and shifts the point of time when the smiths are build. This configuration enables us to produce AIs which are some x% weaker then another AI. <- please give me feedback if the current configurations are suitable or if I should change them
- BuildingListEconomyMinister can handle with medium and low goods
- with the new BuildingListEconomyMinister and the AIMapInformation the AI builds weapon smiths on very small maps like the quick map which it did not before this pull request.
- temples are build as soon as wine is collectable soon
- Ai does not destroy mines anymore. This was an optimization which caused more trouble then effort
- AiDifficultiesTests uses SpezialSumpf.map which is a map some years ago by me which was not on any original CD and has identical start positions to better compare AIs in the tests.


jsettlers.logic/src/main/java/jsettlers/ai/construction/BestBigTempleConstructionPositionFinder.java
```diff
@@ -0,0 +1,24 @@
+package jsettlers.ai.construction;
```
michaelzangl: Copyright missing

jsettlers.logic/src/main/java/jsettlers/ai/economy/BuildingListEconomyMinister.java
```diff
@@ -0,0 +1,416 @@
...
+			weaponsBuildings.add(GOLDMINE);
+			for (int ii = 0; ii < aiMapInformation.getNumberOfGoldMelts(); ii++) {
+				weaponsBuildings.add(GOLDMELT);
+				weaponsBuildings.add(STOCK);
+			}
+		}
+	}
+
+	@Override
+	public int getNumberOfParallelConstructionSides(AiStatistics aiStatistics, byte playerId) {
```
michaelzangl: Sides->Sites

jsettlers.logic/src/main/java/jsettlers/ai/economy/BuildingListEconomyMinister.java
```diff
@@ -0,0 +1,416 @@
...
+					defenceBuildings.add(SMALL_LIVINGHOUSE);
+					defenceBuildings.add(COALMINE);
+					defenceBuildings.add(IRONMINE);
+					defenceBuildings.add(MEDIUM_LIVINGHOUSE);
+					return defenceBuildings;
+				}
+			}
+		}
+		// We are not in danger. We need no defence.
+		return new ArrayList<>();
```
michaelzangl: Does this list need to be modified later?
If not:
- Extract the first one to be a global List. Use `Arrays.asList()` for it.
- Use `Collections.emptyList()` here.

jsettlers.logic/src/main/java/jsettlers/ai/construction/BestBigTempleConstructionPositionFinder.java
codingberlin: Copyright added

jsettlers.logic/src/main/java/jsettlers/ai/economy/BuildingListEconomyMinister.java
codingberlin: Typo fixed.

jsettlers.logic/src/main/java/jsettlers/ai/economy/BuildingListEconomyMinister.java
codingberlin: Introduced static array for defense building and avoid building empty lists by refactoring

andreas-eberle:
I found some strange behavior on the attached map. There, the AI is much worse than it was before. Some points are:

* The performance is realy bad. Before there wasn't such a huge impact caused by the AI.
* The red AI player builds two barracks really quickly but has no food supply at all. 
* The red player builds several fishers as close as possible to each other.
* The blue player build three small temples very quickly after one another. After they were finished, at least one of them was not used for a long time.


[close_8-2016-05-01_13-54-39.zip](https://github.com/jsettlers/settlers-remake/files/254121/close_8-2016-05-01_13-54-39.zip)

codingberlin:
I found the problem:
The new AiMapInformation assumes that regarding the resources the map is symmetric.
On the map is a huge amount of fish, but unfortunately within the yellow players land. The red and blue player would like to build 10 Fisherman to supply their mines but they can't build them on the yellow players land.
The performance issue then was caused by searching for a construction positions of the first fisherman, failing to find one and searching again for all other fishermen.
I fixed the performance issue by remembering the failed construction building types and not searching again within one tick.

I have to figure out how to handle with asymmetric resource amounts on maps.

codingberlin:
I reworked the AiMapInformation that it checks for fish if the fish is available in the default partition or in the own partition to take account that there are asymmetric maps.
I do not do the same for coal and iron because you could occupy the enemy mountains, but you can't occupy the enemy fish because you would rely on fish to produce solders to occupy which is then not possible.

I found a second issue: I did not check the amount of resources and forget, that in all water there is fish but maybe zero fish :)

I started to implement the check but now I have some StackOverflow (some bug in my logic). But now my time is up so I come back to this later.

codingberlin:
I think I fixed all logical issues now but the runtime performance decreased. (On my mac one test failed due to 4700 ms instead of 3000 ms max)

I think I should look into this.

codingberlin:
I was not able to reproduce the timing issue. I assume watching youtube while running the AI tests is not soooooo good for the performance measurement :-)

You can review the PR again.

andreas-eberle:
Another thing: The AI tends to have tons of bowmen and only very very few pikemen and swordsmen (e.g. in the endgame I saw 3000 bowmen and 20 infantrists). 

The result of this is, that often the infantrists die at the front of the fight very quickly. Then, the many bowmen kill all enemies but cannot attack the towers. Because they cannot attack the towers but the AI player needs more infantrists to attack the towers, the bowmen wait and die due to new enemies comming or the bowmens in the towers.

Therefore, although the AI had way enough soldiers to win the fight, it was not able to attack the towers. In the best case, it only loses a lot of time while its waiting for new infantrists to come. In the worst case, it looses a lot of soldiers and aquires no land at all. 

I think it would be better to increase the percentage of infantrists the AI uses. Maybe it should also check if it already has 500 bowmen, it could increase the percentage of infantrists even further.

andreas-eberle:
I have the feeling, that the AI does not expand enough towards stone again. Sometimes it even "dies" because it runs out of stone, while it could have easily expanded towards new stone before (e.g. lower left player on mountain lake).

codingberlin:
The "not enough infantry after attack" and the "land expansion" issues str out of scope of this PR.

1) Must be solved in the ArmyGeneral which was not touched in this PR
2) Will be solved Byte introducing pioneers, which will be my follow up.

Should something be changed for this PR?

codingberlin:
Cause of feedback by ICQ I did a third iteration for this PR.

1) It turned out, that the AI tests ran too long so I solved the issue #436 within this PR (weak enemies will be attacked and defeated). The older assumption that a LooserGeneral and a WinnerGeneral would influence the AIs strength turned out invalid. The only important factor is the number of soldiers. I replaced LooserGeneral and WinnerGeneral by a ConfigurableGeneral. The ConfigurableGeneral attacks when a configurable ratio between weakest enemy solders count and own soldiers count is reached. If the enemy has a threshold count of bowmen the ConfigurableGeneral attacks with bowmen only, to save infanterists. If the enemy is fighted down, the infanterists will be send to occupy the enemy's towers.
For a start I configured the general for the VERY_HARD AI to attack early and VERY_EASY to attack later.

2) The amount of buildings is calculated every tick now. It depends on the occupied land and the default partition divided by the maximal number of players of the map. This keeps the number of buildings low at the start of the game on small maps to reserve place for living houses for the soldiers production but gives the AI the possibility to build a lot more buildings when it defeated another player and occupied its land.

3) The AI detects if its running out of space (called isEndgame in the code) and destroys all but one forester to cut all the trees (crushing the lumberjacks and sawmills when they have no trees anymore) to free space for living houses.

Please review again.

andreas-eberle:
@codingberlin: Thanks for the changes, this makes the AI much better.

From taking a first look, I saw that on small maps (e.g. Vulcano, Small Island ), the AI seems to not produce enought settlers. I will check the code later or tomorrow.

codingberlin:
I observed at Small Islands that the blue player lost all of its soldiers while running in the enemy's land after fighting geologists, so I modified the ```isDanger``` method to check the amount of swordsmen in comparison to the amount of towers which causes that the blue player builds an early weapon smith to compensate the lost soldiers.

In addition I modified the ```automaticLivingHousesEnabled``` to fire when at least one weapon smith was build.

Both changes in combination fixes the lack of settlers on small maps.

andreas-eberle:
One more thing: On nicemap (all players AI_VERY_HARD), the red player produces much more iron melts than weapon smiths. The screenshot below shows the main production area of the red player. The player was never pushed back and so never lost any weaponsmiths. However, the player has 40 ironmelts but only 19 weapon smiths.

unbenannt

andreas-eberle:
This is also reproducible on 6playersfun. After 2:30h the red player (which won the game in that game) had 85 iron melts to 25 weapon smiths.

jsettlers.logic/src/main/java/jsettlers/ai/army/ConfigurableGeneral.java
```diff
@@ -44,61 +44,106 @@
  *
```
andreas-eberle: The last do is fixed by this PR, isn't it?

jsettlers.logic/src/main/java/jsettlers/ai/economy/BuildingListEconomyMinister.java
```diff
@@ -0,0 +1,417 @@
+/*******************************************************************************
+ * Copyright (c) 2015
+ * <p/>
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+ * <p/>
+ * The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+ * <p/>
```
andreas-eberle: please remove the <p/> and update copyright year to 2016.

jsettlers.logic/src/main/java/jsettlers/ai/construction/BestBigTempleConstructionPositionFinder.java
andreas-eberle: please remove <p/>s

jsettlers.logic/src/main/java/jsettlers/ai/construction/BestTempleConstructionPositionFinder.java
```diff
@@ -0,0 +1,48 @@
+/**
+ * ****************************************************************************
+ * Copyright (c) 2015
```
andreas-eberle: 2016 and <p />

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/AiMapInformation.java
```diff
@@ -0,0 +1,200 @@
+/*******************************************************************************
+ * Copyright (c) 2015
```
andreas-eberle: 2016

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/AiMapInformation.java
```diff
@@ -0,0 +1,200 @@
...
+ * enough space to be build on the map. If not it reduces the number of smiths unless it hits a threshold. Then it reduces the number of gold smith,
+ * winegrowers and big temples to a minimum set of this buildings. If again this are too much buildings. It keeps reducing smiths.
+ *
+ * TODO: add information where to find the resources and how far they are to let AI use pioneers instead of towers
+ *
+ * @author codingberlin
+ */
+public class AiMapInformation {
+
+	public static final int GRAS = EResourceType.VALUES.length;
```
andreas-eberle: GRASS with to s. Maybe name it GRASS_INDEX

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/AiMapInformation.java
```diff
@@ -0,0 +1,200 @@
...
+	private static final double FARM_TO_MILL_RATIO = 3F / 1F;
+	private static final double FARM_TO_WATERWORKS_RATIO = 3F / 1F;
+	private static final double FARM_TO_PIG_FARM_RATIO = 3F / 1F;
+	private static final double FARM_TO_SLAUGHTER_RATIO = 6F / 1F;
+	private static final double WEAPON_SMITH_TO_LUMBERJACK_RATIO = 20F / 8F;
+	private static final double LUMBERJACK_TO_SAWMILL_RATIO = 2F / 1F;
+	private static final double LUMBERJACK_TO_FORESTER_RATIO = 2F / 1F;
+	private static final double LUMBERJACK_TO_STONE_CUTTER_RATIO = 8F / 5F;
+	private static final float COAL_MINE_TO_SMITH_RATIO = 1F / 1.8F;
+	private static final float GRAS_TO_LUMBERJACK_RATIO = 1360F;
```
andreas-eberle: grass with two s

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/AiMapInformation.java
```diff
@@ -0,0 +1,200 @@
...
+		buildingCounts[EBuildingType.FISHER.ordinal] = numberOfFisher;
+		int numberOfRemainingWeaponSmiths = Math.max(0, numberOfWeaponSmiths - (int) (numberOfFisher * WEAPON_SMITH_TO_FISHER_HUT_RATIO));
+
+		int numberOfFarms = (int) Math.ceil(numberOfRemainingWeaponSmiths / WEAPON_SMITH_TO_FARM_RATIO);
+			buildingCounts[EBuildingType.FARM.ordinal] = numberOfFarms;
+			buildingCounts[EBuildingType.BAKER.ordinal] = (int) Math.ceil(numberOfFarms / FARM_TO_BAKER_RATIO);
+			buildingCounts[EBuildingType.MILL.ordinal] =  (int) Math.ceil(numberOfFarms / FARM_TO_MILL_RATIO);
+			buildingCounts[EBuildingType.WATERWORKS.ordinal] =  (int) Math.ceil(numberOfFarms / FARM_TO_WATERWORKS_RATIO);
+			buildingCounts[EBuildingType.SLAUGHTERHOUSE.ordinal] =  (int) Math.ceil(numberOfFarms / FARM_TO_SLAUGHTER_RATIO);
+			buildingCounts[EBuildingType.PIG_FARM.ordinal] =  (int) Math.ceil(numberOfFarms / FARM_TO_PIG_FARM_RATIO);
```
andreas-eberle: Apply auto formatter.

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/WhatToDoAi.java
```diff
@@ -359,11 +412,13 @@ private int numberOfAvailableToolsForBuildingType(EBuildingType buildingType) {
 			return Integer.MAX_VALUE;
 		}
 		return aiStatistics.getNumberOfMaterialTypeForPlayer(materialType, playerId)
-				- aiStatistics.getNumberOfNotFinishedBuildingTypesForPlayer(buildingType, playerId)
-				- aiStatistics.getNumberOfUnoccupiedBuildingTypeForPlayer(buildingType, playerId);
```
andreas-eberle: Method AIStatistics.getNumberOfUnoccupiedBuildingTypeForPlayer() is never used any more. Is this ok? From the name I would think that information is required somewhere. If it is ok, please remove that method.

jsettlers.logic/src/main/java/jsettlers/ai/army/ConfigurableGeneral.java
codingberlin: Yes, I deleted the last TODO

jsettlers.logic/src/main/java/jsettlers/ai/economy/BuildingListEconomyMinister.java
codingberlin: done

jsettlers.logic/src/main/java/jsettlers/ai/construction/BestTempleConstructionPositionFinder.java
codingberlin: done

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/AiMapInformation.java
codingberlin: done

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/AiMapInformation.java
codingberlin: done

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/AiMapInformation.java
codingberlin: done

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/AiMapInformation.java
codingberlin: done

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/WhatToDoAi.java
codingberlin: method removed

codingberlin:
The issue with too few weapon smiths is solved now.

andreas-eberle:
@codingberlin: I just let the very hard ai (all players very hard) run on nicemap. I still see the double amount of Iron melts in comparision to the number of weaponsmiths. See screenshots. I saw this on all players. The red player was never pushed back => never lost any buildings.

unbenannt
unbenannt2

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/AiBuildingConstants.java
```diff
@@ -0,0 +1,39 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ *******************************************************************************/
+package jsettlers.ai.highlevel;
+
+/**
+ * @author codingberlin
+ */
+public class AiBuildingConstants {
+	public static final double COAL_MINE_TO_IRONORE_MINE_RATIO = 2F / 1F;
```
andreas-eberle: Why do you use double variables if you calculate with floats?

andreas-eberle:
Could it be that the AI first creates a lot of iron melts when it gets new land and then, when all of them are build, calculates that it hasn't resources for building the same amount of weapon smiths? Because I had the feeling that the problem started, when the AI conquered a larger area. It instantly build a lot of iron melts, but no weapon smiths. Maybe it could help to also consider the iron melts in construction when looking at the ratios.

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/AiMapInformation.java
```diff
@@ -37,58 +48,44 @@
  */
 public class AiMapInformation {
 
-	public static final int GRAS = EResourceType.VALUES.length;
+	public static final int GRASS_INDEX = EResourceType.VALUES.length;
 	private static final double FISH_TO_FISHER_HUTS_RATIO = 80F / 1F;
```
andreas-eberle: Why didn't we move these constants to the new AIBuildingConstants class?

codingberlin: The aim of the BuildingConstants is to be used anywhere (in the AI) where the ratio between two buildings is needed (like in AiMapInformation and in EconomyMinister).
The remaining constants of the AiMapInformation are for the AiMapInformation only (due to performance issues the GRASS_INDEX is used by AiStatistics to fill the AiMapInformation to not run through the whole map twice)

codingberlin:
@andreas-eberle you found another issue.
It was old code which prevented building more than one building of a type if the necessary tool was missing (Hammer - Smith) ... I removed that code because not it should not happen anymore that any AI has no working tools production

This should be really fixed now however I found out, that the easy, hard and very_hard AIs are still very close in their strength so I would like to check if I can downsize the number of building material industry before we merge.

(I have to take a break now, the minor issues (double instead of float) I'll do later)

andreas-eberle:
@codingberlin: Take your time and tell me when you're finished. 

Just so you are aware, I found the two bugs #444 and #443 result in the loss of some infantrists and that some barracks do not produce all possible soldiers. However, it shouldn't have a huge impact on the AI as the bugs only occur rarely. I will have a look at them now.

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/AiBuildingConstants.java
codingberlin: That was not by intention (was created by extracting a magic number).
I changed all to float.

codingberlin:
I introduced a second configuration parameter for the BuildingListEconomyMinister: buildingIndustryFactor

It reduces the amount of lumberjacks, foresters, sawmills and stonecutters to slow down the AI.

On river map the strength is circa:
VERY_HARD: 100%
HARD: 73-88%
EASY: 42-59%
VERY_EASY: 13-20%

The PR can be reviewed again.

andreas-eberle:
Looks much better now. At the moment I have only one more request: I think the AI should build more than one tool smith in the later game. From what I saw, the AI tried to build a lot of buildings and it had the materials to do so. But because it had only one tool smith, it was not able to build enough or any shovels for new diggers or hammers for new bricklayers. But with the large scale of its land, it definitely needed more of them. However, the single tool smith was only able to keep up with the tool requests for picks and so on. 

I think this could also fix the sometimes appearing "missing bearers" situations. Most of the times, the AI wants to construct more living houses, but is limited by its digging and bricklaying capacity.

andreas-eberle:
Btw: Great job with the weapon smiths. On your spezialSumpf map, the winning player had 124 weapon smiths and iron melts!

codingberlin:
I added:
When minimum 16 weapon smiths will be build, the 81st building will be the second tool smith (for VARY HARD this is after the 7th weapon smith)

andreas-eberle:
@codingberlin: Thanks for these great changes!

Issue #431, opened 2016-05-09
Bearers should have a higher selection priority than buildings.


andreas-eberle:
Often when you want to select bearers between some buildings, the buildings get selected instead. However, selecting a building is usually done by directly selecting it, instead of by selecting an area which is used for selecting movables.

PR #432, opened 2016-05-09
Fixes updating selection and selection priorty of settlers vs buildings.

### LINKED ISSUE #430: Selections aren't updated
When you have some soldiers selected and some of them die, the selection stays the same and is not updated.

andreas-eberle:
Fixes #430 and #431.


andreas-eberle:

jsettlers.logic/src/main/java/jsettlers/input/GuiInterface.java
```diff
@@ -107,6 +110,14 @@ public GuiInterface(IMapInterfaceConnector connector, IGameClock clock, ITaskSch
 		this.multiplayer = multiplayer;
 		this.constructionMarksCalculator = new ConstructionMarksThread(grid.getConstructionMarksGrid(), clock, player);
 
+		this.refreshSelectionTimer = new Timer("refreshSelectionTimer");
```
michaelzangl: Please do not use a plain timer for everyting that needs updating.

Use `UIUpdater` instead.

jsettlers.logic/src/main/java/jsettlers/input/GuiInterface.java
```diff
@@ -621,8 +632,17 @@ private void setSelection(SelectionSet selection) {
...
+		if (!currentSelection.isEmpty()) {
+			SelectionSet newSelection = new SelectionSet();
+
+			for (ISelectable selected : currentSelection) {
+				if (selected.isSelected()) {
+					newSelection.add(selected);
+				}
+			}
+
+			setSelection(newSelection);
```
michaelzangl: This is a task for the game logic - not the UI.

You have race conditions in this code.

michaelzangl:
The best thing to do is to notify the UI that the selection changed.

Alternatively, you can let the selection panel or it's components (`SelectionRow`) implement `IUpdateReceiver<SelectionSet>` and use an `UIUpdater` to handle the changes. See the `ToolsPanel` on how to use it. Call a `forceUpdate()` to load the values - that way you can simply move all panel initialisation to the update method.

Mind that buildings (and people) can also change players - settlers need to be deselected then.

You do not need a special case handling for empty sets, simply call `removeAll()` on the root panel of your selection layout.

jsettlers.logic/src/main/java/jsettlers/input/SelectionSet.java
```diff
@@ -61,7 +62,8 @@ public synchronized void add(ISelectable selectable) {
 		if (selectionType.priority < this.selectionType.priority) {
 			return; // selectable is of lower priority
 		} else if (selectionType.priority > this.selectionType.priority) {
-			clear();
+			setSelected(false);
+			this.set = new ArrayList<>();
```
michaelzangl: set.clear() ?

andreas-eberle: what do you mean?

jsettlers.logic/src/main/java/jsettlers/input/GuiInterface.java
andreas-eberle: As this is in the jsettlers.logic project, we shoulnd't use UI stuff.

jsettlers.logic/src/main/java/jsettlers/input/GuiInterface.java
andreas-eberle: It actually is in the game logic.

jsettlers.logic/src/main/java/jsettlers/input/SelectionSet.java
michaelzangl: this.set.clear() does not allocate a new list.

andreas-eberle:
@michaelzangl: Is it ok for you now?

andreas-eberle:
As discussed with @michaelzangl in chat, PR is accepted.

PR #456, opened 2016-05-23
Android preload images v2

michaelzangl:
Improve image preload/cache speed and packing. It should now be more efficient, especially for Android.


andreas-eberle:
@michaelzangl: What is the status on this? Is it still relevant? If so, what needs to be done to get it mergeable?

PR #470, opened 2016-05-26
Added some more maps to the release folder. 

andreas-eberle:


andreas-eberle:
@codingberlin: What do you think about the three maps (apart from SpezialSumpf)?

michaelzangl:
I think we should use a better naming scheme for the file names.

andreas-eberle:
Well, the filenames are not really of interest for the user. He will see a better name in the UI. For now I oriented the filenames on the ones used in the original game (map size-max players-map name). Once the UI displays the map's properties better, we can get rid of the map size-max players prefix. For now, I think it is ok.

@codingberlin: What do you think?

michaelzangl:
But the file name is what we devs use to find out which map is which. I would like to have the same naming scheme for all files in that directory.

andreas-eberle:
@michaelzangl: I changed the naming scheme of the remake maps to 
`<mapsize>-<max playable players>-<name_with_underlines>.<ending>`

I did not change the naming of SpecialSumpf_20, as this is a map downloaded from the settlers mapbase maps designed for the original game by @codingberlin. What do you think?

codingberlin:
The  schema of the map name is not so important for me cause of the search in the GUI.

michaelzangl:
Names are much better now ;-)

PR #479, opened 2016-06-04
Improving AI performance by reducing used search radius when calling AiPositions.getNearestPoint()

andreas-eberle:
Fixes #479.

### DIFF
--- a/jsettlers.logic/src/main/java/jsettlers/ai/construction/BestFisherConstructionPositionFinder.java
+++ b/jsettlers.logic/src/main/java/jsettlers/ai/construction/BestFisherConstructionPositionFinder.java
@@ -40,13 +40,13 @@ public class BestFisherConstructionPositionFinder implements IBestConstructionPo
 	@Override
 	public ShortPoint2D findBestConstructionPosition(AiStatistics aiStatistics, AbstractConstructionMarkableMap constructionMap, byte playerId) {
 		List<ScoredConstructionPosition> scoredConstructionPositions = new ArrayList<ScoredConstructionPosition>();
-		int fishDistance = Integer.MAX_VALUE;
+
 		for (ShortPoint2D point : aiStatistics.getLandForPlayer(playerId)) {
 			if (constructionMap.canConstructAt(point.x, point.y, buildingType, playerId) && !aiStatistics.blocksWorkingAreaOfOtherBuilding(point,
 					playerId, buildingType)) {
-				ShortPoint2D fishPosition = aiStatistics.getNearestResourcePointForPlayer(point, FISH, playerId, fishDistance);
+				ShortPoint2D fishPosition = aiStatistics.getNearestResourcePointForPlayer(point, FISH, playerId, buildingType.getWorkradius());
 				if (fishPosition != null) {
-					fishDistance = point.getOnGridDistTo(fishPosition);
+					int fishDistance = point.getOnGridDistTo(fishPosition);
 					scoredConstructionPositions.add(new ScoredConstructionPosition(point, fishDistance));
 				}
 			}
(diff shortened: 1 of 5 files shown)


andreas-eberle:
@codingberlin: Can you have a look at the changes? There is still some usages with searchRadius=Integer.MaxValue in the BestMilitaryConstructionPositionFinder. Can you have a look if we can improve that too? If it cannot be done so easily, I think we can leave that atm.

jsettlers.logic/src/main/java/jsettlers/ai/construction/BestFisherConstructionPositionFinder.java
```diff
@@ -40,13 +40,13 @@ public BestFisherConstructionPositionFinder(EBuildingType buildingType) {
 	@Override
 	public ShortPoint2D findBestConstructionPosition(AiStatistics aiStatistics, AbstractConstructionMarkableMap constructionMap, byte playerId) {
 		List<ScoredConstructionPosition> scoredConstructionPositions = new ArrayList<ScoredConstructionPosition>();
-		int fishDistance = Integer.MAX_VALUE;
```
codingberlin: Dont remove the outside-loop variable here but instead set the work area radius here because ... (see next comment)

jsettlers.logic/src/main/java/jsettlers/ai/construction/BestFisherConstructionPositionFinder.java
```diff
@@ -40,13 +40,13 @@ public BestFisherConstructionPositionFinder(EBuildingType buildingType) {
...
+
 		for (ShortPoint2D point : aiStatistics.getLandForPlayer(playerId)) {
 			if (constructionMap.canConstructAt(point.x, point.y, buildingType, playerId) && !aiStatistics.blocksWorkingAreaOfOtherBuilding(point,
 					playerId, buildingType)) {
-				ShortPoint2D fishPosition = aiStatistics.getNearestResourcePointForPlayer(point, FISH, playerId, fishDistance);
+				ShortPoint2D fishPosition = aiStatistics.getNearestResourcePointForPlayer(point, FISH, playerId, buildingType.getWorkradius());
 				if (fishPosition != null) {
-					fishDistance = point.getOnGridDistTo(fishPosition);
```
codingberlin: ... here the distance was decreased.
The idea behind that is that if you have found once a fisher which has a close distance (lets say 3) then in the next loop step only fisher will be searched which are even closer which ... (see next comment)

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/WhatToDoAi.java
```diff
@@ -246,7 +240,8 @@ private void destroyBuildings() {
 					taskScheduler.scheduleTask(new DestroyBuildingGuiTask(playerId, lumberJackPosition));
 				}
 			}
-			if ((aiStatistics.getNumberOfBuildingTypeForPlayer(SAWMILL, playerId) * 3 - 2) > aiStatistics.getNumberOfBuildingTypeForPlayer(LUMBERJACK,
+			if ((aiStatistics.getNumberOfBuildingTypeForPlayer(SAWMILL, playerId) * 3 - 2) > aiStatistics.getNumberOfBuildingTypeForPlayer(
+					LUMBERJACK,
 					playerId)) {
```
codingberlin: formatting: playerId could be at the previous line.

codingberlin:
The BestMilitaryConstructionPositionFinder can be left out because I will rework it in the pioneers issue because pioneers will occupy the important resources then and the towers just protect the land against enemies which is a different calculation)

jsettlers.logic/src/main/java/jsettlers/ai/construction/BestFisherConstructionPositionFinder.java
andreas-eberle: I see, i will change that back.

andreas-eberle:
OK, I see that with the BestMilitaryConstructionPositionFinder.

PR #480, opened 2016-06-06
Updated multiplayer server ip to new instance.

andreas-eberle:
As my old server contract was canceled by the provider, I switched to a new instance with the new IP.


justangel:
hi @andreas-eberle if you needed a server, i have my own host on windows (and ubuntu if needed)
static ip address, the server running 24/7, it's free (16gb of RAM, 4 cores), so you do not pay for hosting =)

PR #488, opened 2016-06-26
Ai uses pioneers

codingberlin:
- Ai does not spread the land with towers anymore, it uses two pioneer group.
- the resource pioneer group builds pioneer-streets to missing trees, stones, coal, ironore, goldore and separated partitions from homeland. It also completes stone fields at the border of the players land.
- the broaden pioneer group occupies land which is near the centroid of the players land to broaden the land to get space for more buildings
- towers will be build when another player threats the own land which is not protected by some own tower.
- in mid and late game the AI builds a second small tower, a big tower and even a castle to win time to defeat an attack before the AI is dead by loosing the last tower
- low goods and medium goods get additional medium living houses in the early game to generate enough pioneers to spread the land.

The introduction of pioneers instead of towers fastens the AI so that it produces facto 1,5 more solders after 85 minutes at Halber_Fish map. See the test output after 85 min:

```
Endgame statistic:
Player 0: amountOfProducedSoldiers: 1070, amountOfProducedGold: 16, amountOfProducedMana: 177
```


michaelzangl:
I re-triggered the AI test.

jsettlers.logic/src/main/java/jsettlers/ai/construction/PioneerAi.java
```diff
@@ -0,0 +1,196 @@
...
+		if (!enoughTreesFoundAlready) {
+			ShortPoint2D treeTarget = targetForCuttingBuilding(myBorder, EMapObjectType.TREE_ADULT,
+					EBuildingType.LUMBERJACK, aiStatistics.getTreesForPlayer(playerId), 10);
+			if (treeTarget == null)
+				enoughTreesFoundAlready = true;
+			else
+				return treeTarget;
+		}
+
+		ShortPoint2D target = targetForNearStoneFields();
```
michaelzangl: You can make this cool and object-oriented:
```java
interface ITargetFinder() { ShortPoint2D findTarget(myBorder) }

ITargetFinder[] targetFinders = new ITargetFinder[] {
   /* your coal, fish, ... implementatopn*/ };

for (ITargetFinder targetFinder : targetFinders) {
  // return if it found one
}

File: jsettlers.logic/src/main/java/jsettlers/ai/economy/BuildingListEconomyMinister.java [hunk 2]
```diff
@@ -54,19 +54,19 @@
 	private boolean isMiddleGoodsGame;
 
 	static {
-		EBuildingType[] buildingIndustry = {LUMBERJACK, FORESTER, SAWMILL, STONECUTTER};
+		EBuildingType[] buildingIndustry = { LUMBERJACK, FORESTER, SAWMILL, STONECUTTER };
 		BUILDING_INDUSTRY = Arrays.asList(buildingIndustry);
```
michaelzangl: Use a Set<EBuildingType> here, since the list is not ordered.

Hint: The EnumSet class provides you with really fast enum sets.

jsettlers.logic/src/main/java/jsettlers/ai/economy/BuildingListEconomyMinister.java
```diff
@@ -339,7 +348,7 @@ private void addMinimalBuildingMaterialBuildings() {
 			addIfPossible(STONECUTTER);
 			addIfPossible(LUMBERJACK);
 			addIfPossible(FORESTER);
-			buildingsToBuild.add(TOWER);
+			buildingsToBuild.add(MEDIUM_LIVINGHOUSE);
 			buildingsToBuild.add(MEDIUM_LIVINGHOUSE);
```
michaelzangl: Are you sure it is good to add two medium living houses?

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/AiStatistics.java
```diff
@@ -221,20 +251,53 @@ private void updateMapStatistics() {
...
+	}
+
+	private boolean isBorderOf(short x, short y, byte playerId) {
+		return isIngestibleBy((short) (x + 1), (short) (y + 1), playerId)
+				|| isIngestibleBy((short) (x + 1), (short) (y - 1), playerId)
+				|| isIngestibleBy((short) (x - 1), (short) (y + 1), playerId)
+				|| isIngestibleBy((short) (x - 1), (short) (y - 1), playerId);
+	}
+
+	private boolean isIngestibleBy(short x, short y, byte playerId) {
```
michaelzangl: You can make this method (int, int, byte) to save some casts.

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/AiStatistics.java
```diff
@@ -277,13 +340,27 @@ private void updateFreeLand(short x, short y) {
...
 		}
 		ELandscapeType landscape = landscapeGrid.getLandscapeTypeAt(x, y);
 		if (landscape.isRiver()) {
 			sortedRiversInDefaultPartition.addNoCollission(x, y);
 		}
 	}
 
+	private void updateNearStones(short x, short y) {
+		for (EDirection dir : EDirection.VALUES) {
+			int dx = x + dir.gridDeltaX * NEAR_STONE_DISTANCE;
```
michaelzangl: Feel free to add this method to EDirection:
```java
public short getNextTileX(int x, int steps)
```

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/PioneerGroup.java
```diff
@@ -0,0 +1,106 @@
...
+	private PioneerGroup(List<Integer> pioneerIds, int targetSize) {
+		this.targetSize = targetSize;
+		this.pioneerIds = pioneerIds;
+	}
+
+	public PioneerGroup(int targetSize) {
+		this(new ArrayList<Integer>(targetSize), targetSize);
+	}
+
+	public PioneerGroup(List<Integer> pioneerIds) {
```
michaelzangl: Are you sure you don't want to copy that list?
If yes, please add a note that that list may not be modified after constructing the object.

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/PioneerGroup.java
```diff
@@ -0,0 +1,106 @@
...
+
+		if (newPioneerIds.size() > 0) {
+			taskScheduler.scheduleTask(new ConvertGuiTask(playerId, newPioneerIds, EMovableType.PIONEER));
+		}
+	}
+
+	public PioneerGroup getPioneersWithNoAction() {
+		List<Integer> pioneerIdsWithNoAction = new ArrayList<>(pioneerIds.size());
+		for (Integer pioneerId : pioneerIds) {
+			if (Movable.getMovableByID(pioneerId).getAction() == EMovableAction.NO_ACTION) {
```
michaelzangl: How are you tracking movables that have been killed?

michaelzangl:
Nice patch. It should make gameplay much more interesting.

```
jsettlers.integration.ai.AiDifficultiesIT > veryHardShouldConquerHard FAILED
```
`:jsettlers.logic:test:aiDifficultiesIT` failed, can you have a look at it? Other than that and the minor issues I marked it looks fine and ready to be merged ;-)

jsettlers.logic/src/main/java/jsettlers/ai/economy/BuildingListEconomyMinister.java
codingberlin: Yes I am sure.
First of all with low goods you need more bearers at that time because almost every settler is consumed as a builder, digger or worker.  Tests showed that if you build only a small living house you will need another small living house soon because 10 more settlers are not enough and two small living houses are nearly the same costs like a medium one but for a medium you get 30 instead of 20 settlers. So the first medium living house is for the basic bearers.
The second living house is to occupy the necessary land for the first mines for the tool smith for more lumberjacks and stonecutters. With this second living house you get only up to 30 pioneers of the planned 60. But this is a good compromise between enough pioneers to get enough land fast and costs before the AI starts with the additional lumberjacks and stonecutters.

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/PioneerGroup.java
codingberlin: With this method:
```
public void removeDeadPioneers() {
 		Collection<Integer> idsToRemove = new ArrayList<>(pioneerIds.size());
 		for (Integer pioneerId : pioneerIds) {
 			if (Movable.getMovableByID(pioneerId) == null) {
 				idsToRemove.add(pioneerId);
			}
		}
		pioneerIds.removeAll(idsToRemove);
	}
```
which is called here
```
private void commandPioneers() {
 		if (aiStatistics.getNumberOfTotalBuildingsForPlayer(playerId) < 4 || aiStatistics.getBorderOf(playerId).size() == 0 ||
 				aiStatistics.getEnemiesInTownOf(playerId).size() > 0) {
 			return;
 		}
 
 		resourcePioneers.removeDeadPioneers();
		broadenerPioneers.removeDeadPioneers();

		if (!resourcePioneers.isFull()) {
                    //...
```

codingberlin:
@michaelzangl I will the comments within the next days.
But regarding the AIDifficultiesIT it is the runtime performance of the AI. So this might be a flickering test depending on how busy travis is while running it.

Unfortunately I can't see how much the test is out of performance. (if I can adjust it by 10 ms or if there might be an performance issue e.g. for mobile phones) In the log files there is a hint where maybe we can find it:
```
> There were failing tests. See the report at: file:///home/travis/build/jsettlers/settlers-remake/jsettlers.logic/testRoot/build/reports/tests/index.html
```
Do you know if we can get this file?

michaelzangl:
The files are deleted by travis.

You can try adding this to the main gradle file:
```groovy
    testLogging {
        events "failed"
        exceptionFormat "short"
    }
```

codingberlin:
@michaelzangl 
´´´
test {
    testLogging {
        events "failed"
        exceptionFormat "short"
    }
}
´´´
Did not help to get more output. Do you have another idea how to get the fail message in travis?

jsettlers.logic/src/main/java/jsettlers/ai/construction/PioneerAi.java
codingberlin: done

michaelzangl:
The main project does not have any tests.

Search for the line
```
plugins.withType(JavaPlugin).whenPluginAdded {
```
You should be able to add that block there.

codingberlin:
@michaelzangl 
An additional ```--info``` helped to get the error output.
The AI has a runtime duration median of 53 ms now. The test breaks with 51 ms.
Would you like to check the code to find some performance improvements or is 61 ms ok for the AI and I change the test expectation?
```
script: GRADLE_OPTS='-Xmx600m -Dorg.gradle.jvmargs="-Xmx1500m"' ./gradlew --info $TARGET
```

```
to apply rules:  -> number of measurements: 288, min: 0 ms, average: 13.0 ms, median: 4 ms, max: 259 ms
    to update statistics:  -> number of measurements: 288, min: 32 ms, average: 58.0 ms, median: 53 ms, max: 324 ms

jsettlers.integration.ai.AiDifficultiesIT > veryHardShouldConquerHard STANDARD_ERROR
    Jul 04, 2016 10:40:06 PM jsettlers.integration.ai.AiDifficultiesIT ensureRuntimePerformance
    INFO: to update statistics's median is higher than 50. It was 53.
    Something in the code changed which caused the AI to have a worse runtime performance.
```
```
to apply rules:  -> number of measurements: 288, min: 0 ms, average: 13.0 ms, median: 4 ms, max: 259 ms
    to update statistics:  -> number of measurements: 288, min: 32 ms, average: 58.0 ms, median: 53 ms, max: 324 ms

jsettlers.integration.ai.AiDifficultiesIT > veryHardShouldConquerHard STANDARD_ERROR
    Jul 04, 2016 10:40:06 PM jsettlers.integration.ai.AiDifficultiesIT ensureRuntimePerformance
    INFO: to update statistics's median is higher than 50. It was 53.
    Something in the code changed which caused the AI to have a worse runtime performance.
```

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/AiStatistics.java
```diff
@@ -616,12 +696,16 @@ public AiMapInformation getAiMapInformation() {
 		short partitionIdToBuildOn;
 		IPartitionData materials;
 		AiPositions landToBuildOn;
-		AiPositions borderLandNextToFreeLand;
+		AiPositions border;
```
michaelzangl: final?

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/AiStatistics.java
```diff
@@ -631,16 +715,19 @@ public AiMapInformation getAiMapInformation() {
 		PlayerStatistic() {
 			buildingPositions = new HashMap<EBuildingType, List<ShortPoint2D>>();
 			stones = new AiPositions();
+			stonesNearBy = new AiPositions();
 			trees = new AiPositions();
 			rivers = new AiPositions();
 			landToBuildOn = new AiPositions();
 			enemyTroopsInTown = new AiPositions();
-			borderLandNextToFreeLand = new AiPositions();
+			border = new AiPositions();
```
michaelzangl: You can directly initialize this, you don't need to do it in the constructor.

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/AiStatistics.java
```diff
@@ -650,26 +737,36 @@ public void clearAll() {
...
 			clearIntegerArray(buildingsNumbers);
+			clearLongArray(resourceCount);
 			numberOfNotFinishedBuildings = 0;
 			numberOfTotalBuildings = 0;
 			numberOfNotOccupiedMilitaryBuildings = 0;
 			wineCount = 0;
 			partitionIdToBuildOn = Short.MIN_VALUE;
 		}
 
+		private void clearLongArray(long[] theArray) {
```
michaelzangl: Use `Arrays.fill`

jsettlers.logic/src/test/java/jsettlers/integration/ai/AiDifficultiesIT.java
```diff
@@ -126,10 +133,14 @@ private void holdBattleBetween(EPlayerType expectedWinner, EPlayerType expectedL
 	private void ensureRuntimePerformance(String description, StatisticsStopWatch stopWatch, long median, int max) {
 		System.out.println(description + ": " + stopWatch);
 		if (stopWatch.getMedian() > median) {
+			log.info(description + "'s median is higher than " + median + ". It was " + stopWatch.getMedian() + ".\nSomething in the code changed which "
```
michaelzangl: Extract to local variable.

michaelzangl:
<100ms is OK. I am most concerned with the max peaks. They are over 300ms and thus noticeable by the user.
Can you split that work on multiple ticks to prevent this?

Some more performance hints: `threatenedBorder` should not be an ArrayList<ShortPoint2D>. Use `AiPositions`

`getNearestPoint`: If you only need to find any point, use a method that only scans for that and returns as soon as a point closer to the target was found. Mind that you can square the distance to avoid `sqrt`

`buildingPositions`: Use an `EnumMap` instead. It is faster.

I also added some minor notes to the code ;-)

codingberlin:
@michaelzangl
I can separate the update of the AiStatistics and each calculation of the WhatToDo's in different ticks (which sometimes could cause wrong decisions e.g. if a movable is dead already in the later tick)

How many ticks should be between the operations?

michaelzangl:
This depends on how long each of those steps takes. I'd say something around 10-20 ticks would be enough - it is just to make the game appear more responsive.

codingberlin:
Just a short note: I'm still alive :)
Due to travelling and some construction works in my garden finishing the open issues for this PR is delayed a bit.
But I will come back to this.

andreas-eberle:
@codingberlin: Is it possible to have both AI styles (with towers and with pioneers) in the game? This would allow the user to choose between them and it would make it possible to let them fight each other to see the advantages/disadvantages and detect potential problems.

andreas-eberle:
What I see in the game is, that the AI does not pull back the pioneers when they are attacked. Furthermore, they should be converted back to bearers, when there is no more job for them (all borders are fixed by towers).

jsettlers.logic/src/main/java/jsettlers/ai/economy/BuildingListEconomyMinister.java
codingberlin: done

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/AiStatistics.java
codingberlin: done.

@michaelzangl I decided to implement it with duplication of code:
```java
	public final short getNextTileX(int x) {
		return (short) (x + gridDeltaX);
	}

	public final short getNextTileY(int y) {
		return (short) (y + gridDeltaY);
	}

	public final short getNextTileX(int x, int steps) {
		return (short) (x + gridDeltaX * steps);
	}

	public final short getNextTileY(int y, int steps) {
		return (short) (y + gridDeltaY * steps);
	}
```

rather than delegation:
```java
	public final short getNextTileX(int x) {
		return getNextTileX(x, 1); 
	}

	public final short getNextTileY(int y) {
		return getNextTileY(y, 1);
	}

	public final short getNextTileX(int x, int steps) {
		return (short) (x + gridDeltaX * steps);
	}

	public final short getNextTileY(int y, int steps) {
		return (short) (y + gridDeltaY * steps);
	}
```
Because I think the extra multiplication by 1 could influence performance.
Do you agree or should I use the delegation approach?

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/PioneerGroup.java
codingberlin: fixed

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/AiStatistics.java
codingberlin: fixed

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/AiStatistics.java
codingberlin: done

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/AiStatistics.java
codingberlin: done

jsettlers.logic/src/test/java/jsettlers/integration/ai/AiDifficultiesIT.java
codingberlin: @michaelzangl 
I'm not sure what do you mean. Was it the following change?
```java
if (stopWatch.getMedian() > median) {
			fail(String.format(LOW_PERFORMANCE_FAILURE_MESSAGE, description, "median", median, stopWatch.getMedian()));
		}
		if (stopWatch.getMax() > max) {
			fail(String.format(LOW_PERFORMANCE_FAILURE_MESSAGE, description, "max", max, stopWatch.getMax()));
		}
```

codingberlin:
@andreas-eberle 
To add pioneers to the AI I added a lot of code. On the other hand I was able to delete the old code for the towers in AiStatistics:

```java
private void updateBorderlandNextToFreeLand(PlayerStatistic playerStatistic, short x, short y) {
	for (EDirection dir : EDirection.VALUES) {
			int lx = x + dir.gridDeltaX * BORDER_LAND_WIDTH;
			int ly = y + dir.gridDeltaY * BORDER_LAND_WIDTH;
			if (mainGrid.isInBounds(lx, ly)) {
				if (partitionsGrid.isDefaultPartition(partitionsGrid.getPartitionIdAt(lx, ly))) {
					playerStatistic.borderLandNextToFreeLand.addNoCollission(x, y);
					break;
				}
			}
		}
	}

	public List<ShortPoint2D> getHinterlandMilitaryBuildingPositionsOfPlayer(byte playerId) {
		List<ShortPoint2D> hinterlandMilitaryBuildingPositions = new ArrayList<ShortPoint2D>();
		for (ShortPoint2D militaryBuildingPosition : getBuildingPositionsOfTypesForPlayer(EBuildingType.getMilitaryBuildings(), playerId)) {
			Building militaryBuilding = getBuildingAt(militaryBuildingPosition);
			if (isMilitaryBuildingInHinterland(militaryBuilding, playerId)) {
				hinterlandMilitaryBuildingPositions.add(militaryBuildingPosition);
			}
		}
		return hinterlandMilitaryBuildingPositions;
	}

	private boolean isMilitaryBuildingInHinterland(Building militaryBuilding, byte playerId) {
		for (ShortPoint2D influencedPositions : new MapCircle(militaryBuilding.getPos(), CommonConstants.TOWER_RADIUS + TOWER_RADIUS_OVERLAP)) {
			if (!mainGrid.isInBounds(influencedPositions.x, influencedPositions.y)) {
				continue;
			}
			if (positionIsBorderLandAndIsProtectedOnlyFromOneTower(playerId, influencedPositions) ||
					positionIsOtherPlayersLand(influencedPositions, playerId)) {
				return false;
			}
		}
		return true;
	}

	private boolean positionIsOtherPlayersLand(ShortPoint2D position, byte playerId) {
		return mainGrid.getPartitionsGrid().getPartitionIdAt(position.x, position.y) != playerStatistics[playerId].partitionIdToBuildOn;
	}

	private boolean positionIsBorderLandAndIsProtectedOnlyFromOneTower(byte playerId, ShortPoint2D occupiedPosition) {
		return getBorderLandNextToFreeLandForPlayer(playerId).contains(occupiedPosition)
				&& partitionsGrid.getTowerCountAt(occupiedPosition.x, occupiedPosition.y) == 1;
	}
```

... and the ```BestMilitaryConstructionPositionFinder.java``` changed completely.

Basically it is possible to bring that code back but I see two things:
1) I was happy to remove the code because I save a bit performance for every AI tick with less code
2) The usage of pioneers is definitely better than the usage of towers. If we reintroduce the usage of towers we can only do this for the weaker AIs (e.g VERY_EASY and EASY) because using towers for the stronger AIs would make them significantly weaker.

The only disadvantage of using pioneers I can see is, that you can kill the AI tactical faster when you attack the only tower when their soldiers are far away. But I counteracted against this by adding a second tower, medium towers and castles into the building plan of the AI.

Please take my arguments in account and decide: If I shall reintroduce the towers, I will do it.

codingberlin:
The pioneers will be released now when an enemy is in town or when there is no more land to occupy.

andreas-eberle:
@codingberlin: Sure, I just thought this would be nice if it wouldn't cost much. But I certainly like the performance gain and cleaner code more ;)

codingberlin:
As discussed via ICQ with @andreas-eberle this pull request is feature complete.
The last open issue is the AI performance.

As discussed with @michaelzangl earlier above in this conversation I separated the AI calculations of multiple AIs into different game ticks with the purpose to increase the frame rate when slow performance peaks occur.
Unfortunately the separation I did locally in the code did not remove the small freezes in the gameplay because the slow peaks are caused by the BestFisherConstructionPositionFinder for only one single AI. On the other hand the separation of the AIs to different game ticks brought some disadvantages like movables which moved since the AiStatistics was calculated.
Thats why I do not follow the separation approach anymore.

I digged deep into the code and took time measurements and I identified the BestFisherConstructionPositionFinder and the BestWinegrowerConstructionPositionFinder as performance killers.
To get an idea of the relations here are the measurements of my local computer (On different devices the measurements will look different but the relation should be the same):
- 380 ms to calculate the best fisher construction position
- 42 ms to calculate the best winegrower position
- 4 ms to calculate any of the other buildings construction position

Because the Fisher have factor 100 I looked into that algorithm first.
The problem here is, that the fisher algorithms goes to every players land position and searches for the nearest fish resource each.
(I am not sure but I assume the AiPositions has O(log n))
```every players land * nearest fish resource = O(n) * O(log n) = O(n * log n)```
So we can think about taking every 10. land position to check for a possible fisher but O(n / 10) is still O(n) which will effect us on larger maps or just larger players land.
If we return after the first possible (but not best) fisher position we have still O(n) if no fisher is possible in the whole players land.

I am open if someone has some idea how to deal with that topic.

**Comment by michaelzangl** [comment]
I'd rather prefer to keep it this way for now (with the peeks). We can do such ugly performance optimizations later.

AIPositions has O(n*n) worst case lookup time, you would need to partition the grid (e.g.) QuadBuckets to get away from that bound. I'd personally prefer to use chunks instead since we only need a limited search most times.

To get better results, you can compute your AIStatistics for every 4th coordinate. This reduces the size by a factor of 16 (or by 64, if you use every 8th position) and also reduces lookup time. It does not really matter if the nearest river is 3 blocks closer or not.

You can also pre-compute several values, like the availability of nearby resources (water, wood, fish, ...) and store them if they have not changed. For this, you need to store an invalidation flag for regions that have been changed by the game and then re-compute those regions.

But we can do this in a later PR.

**Comment by andreas-eberle** [comment]
Regarding the AI performance for finding fisher positions: You could calculate if a position can be used to build a fisher or how good it would be for that once at game start. In this calculation, assume the position belongs to the player (the result is the same for every player). At runtime, when you want to check if it works for a certain player, you can combine the precomputed flag with the calculation if a fisher building can be placed at a certain position. This way, you only need to do the expensive calculation once and only do the calculation if the building can be placed at the position during runtime (which you need to do anyways).

**Comment by andreas-eberle** [comment]
I think I saw three bugs:

* I think there might be a bug in the new changes that the AI converts pioneers back if it doesn't need them anymore. In a game I saw, the AI didn't use pioneers after a certain time in the game any more. Although there were a lot of possibilities to use them and there was enemy ground without towers.
* Also, the AI sometimes doesn't send a soldier to a newly built tower. The result is, that the tower isn't occupied (although the AI had spare soldiers) and the enemy can easily take it over or destroy it by building an own tower if no other tower is protecting that position.
* Sometimes the AI builds mines on very bad locations. With bad, I mean that only very few positions of the mine are on the actual mountain; although there was enough space and resources to build several 100% mines, it build a mine that only had one mountain spot.

**Comment by codingberlin** [comment]
1) I implemented the calculation of possible fish positions at the start of the game.
At my computer this decreased the max peaks of calculation time of the ai sometimes by 30%.

2) I fixed the pioneer issue. It was caused by not clearing the shadow pioneer lists while releasing all pioneers in the case of enemy troops in the players land.

3) I can not reproduce the mines issue. The player builds the mines on the position with the most amount resources. Is it possible, that your observed low mines where build either when the other mountain was not occupied or as a late tight mine in between other 100% mines?

4) Regarding the missing swordsmen for occupying towers I debugged that the AI sends gui tasks for swordsmen to go to the tower but they never follow the task under the condition that currently an attack is running. This means that the same swordsmen got 2 orders. The first is to attack the enemy. The second is to go to the tower. I thought because of the order the swordsmen would go to the tower. @andreas-eberle am I wrong here? Do I need to exclude the tower swordsmen from the attack swordsmen?

**Comment by andreas-eberle** [comment]
Thanks for your update! Regarding

3) If I see the issue again / can reproduce it, I will open a new ticket. I only saw that once yet.

4) Soldiers currently have no priorties for orders or something like that. Therefore if you give a soldier the order to go somewhere and then give him another order, he will follow the last order. However, if a soldier encounters an enemy on his way to the tower, he will engange him, as there is currently no "just go there and don't care about anything else"-mode. 

Only when a tower that is currently needing a soldier finds one and "orders" him to go into the tower, the soldier will always follow that order and do nothing else.

Therefore, I think this can only be fixed by introducing a new goto option for movables. That option is needed anyways to be able to move soldiers away from a fight without spamming go tos.

**Comment by codingberlin** [comment]
I fixed the swordsmen to occupy tower issue.
From my side this is ready to be merged now.

**Comment by andreas-eberle** [comment]
At the moment my computer is broken and in repairs for approx 1 to 3 more weeks. Sorry for the delay...

**Comment by andreas-eberle** [comment]
Regarding the bad mines: When playing mountain_lake with all players being very hard AIs, the red player expands with pioneers towards the mountain area. Then, the AI builds the first mine on the first spot in can build a mine. Depending on the timing, this can be a very bad mine with only one mining spot.

Another thing I saw on the same map: Some pioneers are expanding towards the mountains. However, they expand towards the snowy mountains where there cannot be built any mines, instead of expanding towards the mountain that can be used. Thus, the red player loses a lot of time expanding towards useless mountains.

Anyways: I like how good that all works already!

**Comment by 0815-user** [comment]
Hi,
First of all:  Thanks a lot for the nice project!
I did some testing with all bots “very hard”.
Here are a few things I saw:
1.	I observed at “mountain lake” that the red player expands to a mountain which is unnecessary far away. So he loses time and has no chance against the others. 
2.	Pioneers do not focus on occupying unprotected land from other players to gain from its goods. 
3.	At “Nice Map” I saw sometimes the green player loses its last tower. But the remaining land with its infrastructure creates new soldiers which conquer towers. So green gets strong again. Just an idea: Maybe pioneers can focus on destroying remaining barracks.
4.	It looks like there is a strategy to connect the land of a player. But maybe it would help to give this strategy a higher priory.
5.	It looks like the expansion rate gets reduced after some time. Maybe the philosophy of continuously expanding would be helpful. Later in the game the player has usually more free settlers, so the resources would even allow an increased expansion rate. In this way the bot can get more resources, open parallel paths to reduce traffic jams and even without an extra logic it would occupy unprotected land from other players.

But anyway: 
Great work I am impressed!

**Comment by codingberlin** [comment]
@andreas-eberle 
1) I added a filter which prevents pioneers  and geologists from seeking out very thin resources like between snow and gras.
2) Convert all pioneers back when anywhere an attacker is in the land is the pareto solution: It saves the AI (not running out of settlers) and it saves runtime performance (because of no calculations from where the enemy comes from and if there are more then one enemy and so forth) - An optimization for this point is nothing for the first pioneer iteration.
3) I implemented the check for blockedPartition but unfortunately the glidge at NiceMap still occurs. Nevertheless this fix prevents the bug in other situations. (I recognized special circumstances for orange at NiceMap because of the river tiles which are in the water tiles at this position)

@0815-user 
1) The red position is a bit unbalanced. This are really the nearest resources for red.
2) Thats true. An optimization could take place in a second iteration for the pioneer AI
3) This is connected with (2). When pioneers take the enemies land, they could focus on barracks first
4) The priority of connecting the land is 4 of 9. That's pretty high. The higher 3 focus on trees and stones. This is because of a split land situation when an enemy nearly destroyed the AI. The most important thing then is to get back lumberjacks and stonecutters to be able to rebuild the later connected land.
5) Because of runtime performance the pioneers search their new targets within a maximum distance. With my last commit I added a strategy witch reallocates the search center in the case no pioneer target could be found. I noticed that in the later game there could be more pioneers but on the other hand the AI has so many land that there is enough space to build buildings so that more pioneers are not necessary. I ran the automated tests with different amounts of pioneers and choose the number of pioneers which led to the most soldier production in the end.

@andreas-eberle 
If there is no major bug now this was my last commit.

**Comment by 0815-user** [comment]
1)	At “mountain lake” the red player expands to the first mountain in the north which has coal and iron. It places an iron mine but instead of doing the same for coal it further expands to the mountain behind the snowy part of the mountain and builds a coal mine there. So the coal mine is unnecessary far away.

File: jsettlers.logic/src/main/java/jsettlers/ai/highlevel/AiStatistics.java [hunk 12]
```diff
@@ -193,8 +236,33 @@ private void updateMapStatistics() {
...
+								if (fishMapInformationPlayerId == aiMapInformation.resourceAndGrassCount.length - 1) {
+									fishMapInformationPlayerId = mapInformationPlayerIdOfPosition(x, (short) (y + 3));
+									if (fishMapInformationPlayerId == aiMapInformation.resourceAndGrassCount.length - 1) {
+										fishMapInformationPlayerId = mapInformationPlayerIdOfPosition(x, (short) (y - 3));
+									}
+								}
+							}
+						}
+						aiMapInformation.resourceAndGrassCount[fishMapInformationPlayerId][resourceType.ordinal]++;
+						if (fishMapInformationPlayerId != aiMapInformation.resourceAndGrassCount.length - 1) {
```
andreas-eberle: Why are you using this complicated value as guard instead of -1 or MIN_VALUE?

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/AiStatistics.java
```diff
@@ -277,13 +381,27 @@ private void updateFreeLand(short x, short y) {
...
 	}
 
+	private void updateNearStones(short x, short y) {
+		for (EDirection dir : EDirection.VALUES) {
+			int dx = dir.getNextTileX(x, NEAR_STONE_DISTANCE);
+			int dy = dir.getNextTileY(y, NEAR_STONE_DISTANCE);
+			if (mainGrid.isInBounds(dx, dy)) {
+				byte playerId = partitionsGrid.getPlayerIdAt(dx, dy);
+				if (playerId != -1  && hasPlayersBlockedPartition(playerId, x, y)) {
+					playerStatistics[playerId].stonesNearBy.addNoCollission(x, y);
```
andreas-eberle: Am I correct that you add any stones in that area multiple times? I mean, I stone has a ground area larger 1 and therefore will be added multiple times. Might this cause problems? Or is it just not optimal?

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/WhatToDoAi.java
```diff
@@ -266,27 +290,53 @@ private boolean isWoodJam() {
...
+		List<ShortPoint2D> threatenedBorder = aiStatistics.threatenedBorderOf(playerId);
+		if (threatenedBorder.size() == 0) {
+			return false;
+		}
+
+		ShortPoint2D position = bestConstructionPositionFinderFactory
+				.getBorderDefenceConstructionPosition(threatenedBorder)
+				.findBestConstructionPosition(aiStatistics, mainGrid.getConstructionMarksGrid(), playerId);
+		if (position != null) {
+			taskScheduler.scheduleTask(new ConstructBuildingTask(EGuiAction.BUILD, playerId, position, TOWER));
```
andreas-eberle: @codingberlin: What do you think about increasing the tower's build priority. The tower is build because of a border that needs to be defended. Therefore I think it could help to speed up this build by increaing it's priority.

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/WhatToDoAi.java
```diff
@@ -361,15 +411,70 @@ private void addBuildingCountToBuildingPlan(EBuildingType buildingType, Map<EBui
...
+
+		if (!resourcePioneers.isFull()) {
+			fill(resourcePioneers);
+			setNewTargetForResourcePioneers();
+		} else {
+			setNewTargetForResourcePioneers();
+			if (!broadenerPioneers.isFull()) {
+				fill(broadenerPioneers);
+			}
+			if (broadenerPioneers.isNotEmpty()) {
```
andreas-eberle: This whole if should be moved to a method like setNewTargetForResourcePioneers()

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/pioneers/target/LumberJackTargetFinder.java
```diff
@@ -0,0 +1,48 @@
...
+
+	@Override
+	public ShortPoint2D findTarget(AiPositions playerBorder, ShortPoint2D center) {
+		if (disabled) {
+			return null;
+		}
+
+		ShortPoint2D target = findTarget(playerBorder, center, aiStatistics.getTreesForPlayer(playerId).size());
+		if (target == null) {
+			disabled = true;
```
andreas-eberle: What exactly is the idea behind disabling this target finder? E.g. the stone target finder is never disabled. Is it because of performance?

andreas-eberle:
Apart from my questions, the code looks good to me. Thanks already for this great contribution!

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/AiStatistics.java
codingberlin: ```MapInformation.resourceAndGrassCount``` is an Array. Because there is no index -1 I use playerNumber + 1 as index for the default partition.

However I inverted the if statement to make it more readable.

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/AiStatistics.java
```diff
@@ -221,22 +289,58 @@ private void updateMapStatistics() {
...
 		}
 	}
 
+	private int mapInformationPlayerIdOfPosition(short x, short y) {
+		if (!mainGrid.isInBounds(x, y)) {
+			return aiMapInformation.resourceAndGrassCount.length - 1;
+		}
+
+		byte playerId = mainGrid.getPartitionsGrid().getPlayerIdAt(x, y);
+		if (playerId == -1) {
```
codingberlin: see answer above

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/pioneers/target/LumberJackTargetFinder.java
codingberlin: I changed the behaviour of the code and tried to write the intention into an extracted method name. Is it clear now or should I add a WHY-comment into the code?

```
@Override
public ShortPoint2D findTarget(AiPositions playerBorder, ShortPoint2D center) {
	if (enoughTreesAreGuaranteed()) {
		return null;
	}
	return findTarget(playerBorder, center, aiStatistics.getTreesForPlayer(playerId).size());
}
private boolean enoughTreesAreGuaranteed() {
	return aiStatistics.getNumberOfBuildingTypeForPlayer(EBuildingType.FORESTER, playerId) >= 2;
}
```

(The stones won't grow back so you need to occupy more forever)

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/AiStatistics.java
codingberlin: I was not aware that stones have an area. So yes: This is not optimal.

@andreas-eberle I don't know what needs to be done to achieve adding stones only once. Could you advice me here please?

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/WhatToDoAi.java
codingberlin: That ***may*** be a good idea.

Advantage:
- The border is safe faster.

Disadvantage:
- In some cases the economy is more important than the border. If the border is far away I prefer to produce more soldiers to kill my enemy anyway

Possible compromis:
- enable priority if buildings are some kind of near by

Because this is on the one hand an optimisation and on the other hand an optimisation with further investment to figure out the right optimum I would keep this out of this PR

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/WhatToDoAi.java
```diff
@@ -361,15 +411,70 @@ private void addBuildingCountToBuildingPlan(EBuildingType buildingType, Map<EBui
...
+			//pioneers which can not be converted shall walk into player's land to be converted the next tic
+			taskScheduler.scheduleTask(new MoveToGuiTask(playerId, aiStatistics.getPositionOfPartition(playerId), pioneerIds));
+		}
+	}
+
+	private void sendOutPioneers() {
+		resourcePioneers.removeDeadPioneers();
+		broadenerPioneers.removeDeadPioneers();
+
+		if (!resourcePioneers.isFull()) {
```
codingberlin: If I would do this the code would become more complex because of the asynchronous execution of the AI orders.

The Ids of the bearers must be remembered to not double convert a bearer into a pioneer for both groups.

I think the algorithm is better than using no pioneers at all. So we should merge the PR into the codebase and cover optimisations later following the principle to do the Must Haves first and the Nice to Haves later.

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/WhatToDoAi.java
codingberlin: done

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/AiStatistics.java
andreas-eberle: ok

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/pioneers/target/LumberJackTargetFinder.java
andreas-eberle: I got it now. Thx.

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/WhatToDoAi.java
andreas-eberle: I see your point with the asynchronous exection.

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/WhatToDoAi.java
andreas-eberle: I agree. This should be part of another PR.

jsettlers.logic/src/main/java/jsettlers/ai/highlevel/AiStatistics.java
andreas-eberle: In fact, I was wrong. For every stone, only a single map object is placed on the map. However, an area is blocked and protected for that stone. So this should be fine.

andreas-eberle:
Thanks for these changes!

PR #492, opened 2016-08-14
Providing continous builds of master and PRs in branch gh-pages of jsettlers/settlers-nightlies repository

andreas-eberle:
This PR provides the scripts for automatic deployment of the release target's results to the repositiory https://github.com/jsettlers/settlers-nightlies

Currently, every commit on a PR and the master is build and deployed if the complete build was successful.


michaelzangl:
Is there any specific reason you are not using gradle for distribution?

I would suggest to add a new gradle task. We already use the jgit plugin in gradle, it can handle this (if you do not want to work with the command line ;-) ). Or we can use an [Exec](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.Exec.html) task.

andreas-eberle:
This is the way I found here: https://gist.github.com/domenic/ec8b0fc8ab45f39403dd. 

So no, there is no special reason for it. As long as we use the Travis environment variables to decode the encrypted deploy key.

andreas-eberle:

.travis.yml
```diff
@@ -53,8 +44,12 @@ env:
...
-  # Test release generation
+  
   - TARGET="release"
 
+  
 script: GRADLE_OPTS='-Xmx600m -Dorg.gradle.jvmargs="-Xmx1500m"' ./gradlew $TARGET
+
+
+after_success:
+- ./deploy.sh
```
michaelzangl: Won't this run for every build in our matrix?

deploy.sh
```diff
@@ -0,0 +1,88 @@
+#!/bin/bash
+
+set -e # Exit with nonzero exit code if anything fails
+
+TARGET_BRANCH="gh-pages"
+
+function doCompile {
+  GRADLE_OPTS='-Xmx600m -Dorg.gradle.jvmargs="-Xmx1500m"' ./gradlew release
```
michaelzangl: You can simply add a release target to the build matrix. So this step is not required.

deploy.sh
```diff
@@ -0,0 +1,88 @@
...
+
+# Commits to other branches than master shouldn't be deployed
+if [ "$TRAVIS_PULL_REQUEST" == "false" -a -z "$TRAVIS_TAG" -a "$TRAVIS_BRANCH" != "master" ]; then
+    echo "This is the build of a branch => Skipping deploy;"
+    exit 0
+fi
+
+
+if [ -z "$encrypted_af9c5a2dd85c_key" ]; then 
+	echo "No encryption key for deploy provided => Skipping deplo;"
```
michaelzangl: ...y

michaelzangl:
The compile phase does not seem to be required - we invoke gradle just before the script is run, so we should not need to invoke gradle again. We should also ensure that the script is only run for one build.

You can achive this easily by adding a new :githubDeploy target to the root `build.gradle`. Use an exec task that runs the deploy script and depends on `release`. Then remove the gradlew line from the deploy script.

I can have a look at it in about one week (after the GSoC deadline ;-))

.travis.yml
andreas-eberle: No, `after_success` is only run once after all the matrix elements have successfully been build. What you mean is `after_script`.

See https://docs.travis-ci.com/user/customizing-the-build/

deploy.sh
andreas-eberle: As the `after_success` job is run independently of the build steps (see my comment before), I need to build it here, as I cannot garantee the build that runs before the deploy script.

andreas-eberle:
I will have a look at changing it to a gradle task.

.travis.yml
michaelzangl: `after_success` is run in between of `script` and `after_script`. Have a look at the log output, it is called for every build you do ;-)

andreas-eberle: You're right, I misunderstood the documentation... I will have a look on changing it to  a gradle task.

andreas-eberle:
I changed it to use a gradle task. @michaelzangl: is it ok for you?

michaelzangl:
Looks fine to me.

PR #495, opened 2016-08-23
Enhancement add retrolambda and streamsupport to enable usage of lambdas and streams in all subprojects

### LINKED ISSUE #494: Use retrolambda for Android to be able to use lambdas in platform crossing project parts

andreas-eberle:
fixes #494.


michaelzangl:
Can you test if travis notifies us of e.g. `Stream` usage in any project?

andreas-eberle:
The current Android build chain cannot detect this. So we have the danger of adding code that breaks at runtime...

The only way we could get around this is with the new Jack&Jill build chain. I tested it already and that buildchain detects such problems and allows most java8 features (like stream API). However, it is still in a beta status at the moment. Furthermore, Working with lambdas in the java subprojects currently has a bug when you use interfaces of the standard java library...

* So we are left to three options: Wait with lambdas and Stream API until the Jack&Jill build chain is finished
* Use the current build chain and manually make sure we don't use forbidden stuff
* Use the beta Jack&Jill with the stuff that currently works

michaelzangl:
> Use the current build chain and manually make sure we don't use forbidden stuff

Can we detect this? What error do we get if we do something forbidden? I have no problem starting an emulator in Travis if we need to.

andreas-eberle:
I changed the gradle files to use the new Jack&Jill build chain which includes retrolambda. This build chain also detects if we use Java8 Streams in any project included in our Android projects.

Furthermore, I added the streamsupport library to add a backported Java8 streams support. See https://github.com/streamsupport/streamsupport. 

All in all, this change allows us to use streams and lambdas almost like it would be with real Java8 in all projects. Only the package namings of some stream classes differ. 

However, at the moment I'm experiencing three IDE bugs:
1. The new Android Studio doesn't package the resources correctly making it impossible to run the desktop version of JSettlers out of Android Studio.
2. Android Studio marks all uses of the standard Java8 stream API in red. Even the ones in the swing projects that are no Android projects and aren't included by Android projects.
3. IntelliJ IDEA cannot run the Android build on Windows because of a bug that causes Jill to crash as it cannot delete a temporary file. This was fixed in Android Studio a while ago but not in IntelliJ...

I will try to find some solutions for those. @michaelzangl: Can you test this on your computer?

andreas-eberle:
It seams that Jack&Jill have this bug for a while now: https://code.google.com/p/android/issues/detail?id=223700

andreas-eberle:
@michaelzangl: I switched back to retrolambda again, as jack&jill aren't stable atm. Furthermore, I added an android instrumentation test to ensure nobody uses the wrong stream API (if you use streams, always use the ones in the java8 packages comming from the streamsupport library).

Can you have a look at the changes?

andreas-eberle:
I also added a [wiki page](https://github.com/jsettlers/settlers-remake/wiki/Frameworks-&-Java8) linking the two frameworks and giving a short explanation.

michaelzangl:
Seems fine to me - although I did not test it.

PR #529, opened 2017-01-06
Enhancement improve performance of AI and game

### LINKED ISSUE #528: Improve AI startup performance

andreas-eberle:
fixes #528


jsettlers.logic/src/main/java/jsettlers/logic/map/grid/objects/MapObjectsManager.java
```diff
@@ -400,18 +401,14 @@ public boolean pushMaterial(short x, short y, EMaterialType materialType) {
...
+		return HexGridArea.iterate(x, y, 0, 200, (currX, currY, radius) -> {
 			if (grid.isInBounds(currX, currY) && canForcePushMaterial(currX, currY, materialType)) {
 				pushMaterial(currX, currY, materialType);
-				return position;
+				return Optional.of(new ShortPoint2D(currX, currY));
 			}
-		}
-		return null;
+			return Optional.empty();
+		}).orElseGet(() -> null);
```
michaelzangl: `orElse(null)` ;-)

jsettlers.logic/src/main/java/jsettlers/logic/movable/strategies/specialists/PioneerStrategy.java
```diff
@@ -87,22 +89,22 @@ private void findWorkablePosition() {
 		}
 	}
 
-	private final EDirection getCloseForeignTile() {
-		EDirection bestNeighbourDir = null;
-		double bestNeighbourDistance = Double.MAX_VALUE; // distance from start point
+	private EDirection getCloseForeignTile() {
+		EDirection[] bestNeighbourDir = new EDirection[1];
```
michaelzangl: You can use a custom Class here.

You can add the visiting method to that object to make the code clean and save some object creations.

jsettlers.logic/src/main/java/jsettlers/logic/movable/strategies/specialists/PioneerStrategy.java
```diff
@@ -87,22 +89,22 @@ private void findWorkablePosition() {
...
-
-			if (super.isValidPosition(satelitePos) && canWorkOnPos(satelitePos)) {
-				double distance = Math.hypot(satelitePos.x - centerPos.x, satelitePos.y - centerPos.y);
-				if (distance < bestNeighbourDistance) {
-					bestNeighbourDistance = distance;
-					bestNeighbourDir = EDirection.getApproxDirection(position, satelitePos);
+		HexGridArea.iterate(position.x, position.y, 1, 6, visitor(((currX, currY, radius) -> {
+			ShortPoint2D currPosition = new ShortPoint2D(currX, currY);
+			if (super.isValidPosition(currPosition) && canWorkOnPos(currPosition)) {
+				double distance = Math.hypot(currX - centerPos.x, currY - centerPos.y);
```
michaelzangl: (1) This is not the "real" distance.
(2) `Math.hypot(a, b)` is slow, much slower than `Math.sqrt(a*a + b*b)`. Use the squared distance to avoid both of them.

jsettlers.logic/src/main/java/jsettlers/logic/map/grid/objects/MapObjectsManager.java
```diff
@@ -400,18 +401,14 @@ public boolean pushMaterial(short x, short y, EMaterialType materialType) {
...
 	}
 
-	public ShortPoint2D pushMaterialForced(short x, short y, EMaterialType materialType) {
-		IMapArea mapArea = new HexGridArea(x, y, 0, 200);
-
-		for (ShortPoint2D position : mapArea) {
-			short currX = position.x;
-			short currY = position.y;
+	public ShortPoint2D pushMaterialForced(int x, int y, EMaterialType materialType) {
+		return HexGridArea.iterate(x, y, 0, 200, (currX, currY, radius) -> {
```
michaelzangl: Have you tested the speed improvement of this?

It may be sufficient to replace `return position;` with `return new ShortPoint2D(currX, currY)` - this allows the JIT to inline the loop object.

jsettlers.logic/src/main/java/jsettlers/algorithms/distances/DistancesCalculationAlgorithm.java
```diff
@@ -0,0 +1,83 @@
...
+
+	public static BitSet calculatePositionsInDistance(int width, int height, IBooleanCoordinateValueProvider provider, int maxDistance) {
+		int area = width * height;
+		BitSet inDistance = new BitSet(area);
+		BitSet next = new BitSet(area);
+		int distance = 0;
+
+		setupInitial(width, height, provider, inDistance, next);
+
+		while (true) {
```
michaelzangl: for(distance = 1, distance <= maxDistance; distance++) ?

jsettlers.common/src/main/java/jsettlers/common/utils/interfaces/IBooleanCoordinateValueProvider.java
```diff
@@ -0,0 +1,22 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ *******************************************************************************/
+package jsettlers.common.utils.interfaces;
+
+/**
+ * Created by Andreas Eberle on 06.01.2017.
+ */
+public interface IBooleanCoordinateValueProvider {
```
michaelzangl: `ICoordinatePredicate`? (... `extends Predicate<ShortPoint2D>`)

jsettlers.common/src/main/java/jsettlers/common/utils/interfaces/IBooleanCoordinateValueProvider.java
```diff
@@ -0,0 +1,22 @@
...
+	boolean getValue(int x, int y);
```
michaelzangl: `boolean test(int x, int y)`?

jsettlers.common/src/main/java/jsettlers/common/utils/debug/DebugImagesHelper.java
```diff
@@ -0,0 +1,84 @@
...
+			File debugFolder = new File(DEBUG_PATH);
+			deleteRecursive(debugFolder);
+
+			debugFolder.mkdirs();
+		}
+	}
+
+	private static void deleteRecursive(File file) {
+		if (file.isDirectory()) {
+			for (File child : file.listFiles()) {
```
michaelzangl: Stream.of(file.listFiles()).forEach(DebugImagesHelper::deleteRecursive)

jsettlers.common/src/main/java/jsettlers/common/utils/interfaces/ICoordinateValueProvider.java
```diff
@@ -0,0 +1,22 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ *******************************************************************************/
+package jsettlers.common.utils.interfaces;
+
+/**
+ * Created by Andreas Eberle on 06.01.2017.
+ */
+public interface ICoordinateValueProvider<T> {
```
michaelzangl: `IBinaryIntFunction<R>`?

jsettlers.common/src/main/java/jsettlers/common/map/shapes/HexGridArea.java
```diff
@@ -152,10 +158,15 @@ public static void iterate(int cX, int cY, int startRadius, int maxRadius, ICoor
 					x += DIRECTION_INCREASE_X[direction];
 					y += DIRECTION_INCREASE_Y[direction];
 
-					consumer.consume(x, y, radius);
+					T result = visitor.visit(x, y, radius);
+					if (result != null) {
+						return result;
```
michaelzangl: We could return an `Optional` here.

michaelzangl: ... forget it, did not see this was outdated ;-)

jsettlers.logic/src/main/java/jsettlers/logic/map/grid/objects/MapObjectsManager.java
codingberlin: You should avoid `orElse(null)` whenever possible. This is a code smell.

Evaluate the Optional as late as possible. (Except you recover the absence of the value in the middle with a really good recover value)
If your Interfaces provide Optional instead of the possibility of null you can omit the null checks without the runtime possibility of NullPointerExceptions.
I try to show you what I mean with the following two code examples:

```
class Foo {

        public static void main(String[] args) {
                System.out.println("Hallo");
                System.out.println(third());
        }

        public static Integer third() {
                Integer secondResult = second();
                if (secondResult == null)
                        return null;
                return secondResult * 2;
        }

        public static Integer second() {
                Integer firstResult = first();
                if (firstResult == null)
                        return null;
        return firstResult + 15;

        }

        public static Integer first() {
                return null;
        }
}
```

```
import java.util.Optional;

class Bar {

        public static void main(String[] args) {
                System.out.println("Hallo");
                // Here you can decide what to do if this is empty. Recover it? End the program? Ignore it. What you like
                System.out.println(third().orElse(42));
        }

        public static Optional<Integer> third() {
                return second().map(second -> second * 2);
        }

        public static Optional<Integer> second() {
                return first().map(first -> first + 15);
        }

        public static Optional<Integer> first() {
                return Optional.empty();
        }
}
```

I can imagine only two exceptions when using `orElse(null)` is allowed:
1) If you provide code which is used by other code which is not under your control thus you can't change the interface to Optional<T>
2) If you refactor step by step and seek for small commits your can introduce Optional<T> step by step by using `orElse(null)` in the meantime. But your goal at the end should be to avoid `orElse(null)`.

jsettlers.logic/src/main/java/jsettlers/logic/map/grid/objects/MapObjectsManager.java
andreas-eberle: How do you mean? What should the JIT be able to inline and to where? 

My plan is to completely remove the iterators, as they create large amounts of objects. That's the main reason, why I replaced this code here. As this method is not called very often, speed doesn't matter that much, here. However, I don't want to have the iterators and the iterate methods in parallel, as they tend to have duplicated, complicated code.

jsettlers.logic/src/main/java/jsettlers/logic/map/grid/objects/MapObjectsManager.java
andreas-eberle: @michaelzangl: Thanks for the hint. I must have overlooked that.
@codingberlin: I know. Actually I don't want to keep it this way. But at the moment my focus is on optimizing speed and getting away from these iterators creating tons of objects. So it's your second point that I see here at the moment. I just don't want to change all APIs with this change.

jsettlers.logic/src/main/java/jsettlers/logic/movable/strategies/specialists/PioneerStrategy.java
andreas-eberle: I had a look for all the hypots in the code and replaced them either with the better implementation or the ShortPoint2D.getOnGridDist().

andreas-eberle:
Just as a note, this change is not finished yet. I created the PR already to let others take a look, but I still want to adapt more of our Area classes.

andreas-eberle:
@michaelzangl @codingberlin: I think that's it for the moment. Can one of you have a look at the changes? 

Two things:
* The changes should have no effect on the game at all. The auto replay references have only been updated due to naming changes / structural changes. I updated those files in the regarding commits. 
* IntelliJ changed the order of the imports in several classes, although I did not change the formatter settings and have the settings as it is shown in the tutorial in the wiki. Therefore I think IntelliJ might have changed its ordering due to an update or so. So I would say we just adapt to the new ordering.

jsettlers.common/src/main/java/jsettlers/common/utils/coordinates/ICoordinateConsumer.java
```diff
@@ -0,0 +1,24 @@
...
+ * DEALINGS IN THE SOFTWARE.
+ *******************************************************************************/
+package jsettlers.common.utils.coordinates;
+
+import java.io.Serializable;
+
+/**
+ * Created by Andreas Eberle on 06.01.2017.
+ */
+public interface ICoordinateConsumer extends Serializable {
```
michaelzangl: Do we really need to enforce ICoordinateConsumers to be serializable? (same for some other interfaces)

jsettlers.graphics/src/main/java/jsettlers/graphics/map/MapContent.java
```diff
@@ -515,57 +508,13 @@ private void drawTile(int x, int y) {
...
-					0,
-					0,
-			};
-
-			for (ShortPoint2D pos : tiles) {
-				short x = pos.x;
-				short y = pos.y;
+		GLDrawContext gl = this.context.getGl();
+
+		float[] shape = new float[] { 0, 4, .5f, 0, 0, -3, 2, .5f, 0, 0, -3, -2, .5f, 0, 0, 0, -4, .5f, 0, 0, 0, -4, .5f, 0, 0, 3, -2, .5f, 0, 0, 3,
```
michaelzangl: This list was once broken every 5 items... Autoformat :-(

jsettlers.logic/src/main/java/jsettlers/logic/map/grid/partition/PartitionsGrid.java
```diff
@@ -49,6 +42,14 @@
 import jsettlers.logic.player.PlayerSetting;
 import jsettlers.logic.player.Team;
 
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.io.Serializable;
+import java.util.*;
```
michaelzangl: We should not use * imports.

andreas-eberle: Why?

andreas-eberle: Thx, just had to find out my IntelliJ forgot the auto formatter code style again...

PR #532, opened 2017-01-09
Android restructure to allow rotation and game running in background

Tom-Pratt:
Most of the UI has been replaced, a MainActivity shows pre game menus. GameActivity shows loading and gameplay. Game objects are held in an Android service called GameService to allow game to run in background. The jsettlers.graphics.androidui module still contains some unused files and there is general tidying up to be done when more detail is implemented.

PR #536, opened 2017-01-14
Android updates v2

Sylapse:
Hi guys,

I haven't had a go at saving or network games yet so you don't need to try and merge this or look through it in detail.

I did move lots of things around and implement more menus though and thought the code was at a good stage to have a record of before it's ready for merge.

Main thing that changed was breaking up the implementation of IControls. As more and more menus needed access to different parts of the controls I decided to break it into:

- AndroidControls which implements IControls and communicates with the core logic.
- ControlsAdapter, sits in between AndroidControls and the UI, the UI code only ever uses this or its interfaces to communicate with the game.

AndroidControls and ControlsAdapter talk to eachother when passing information between UI and core logic. ControlsAdapter is separated into several interfaces providing lower level information about Actions, Selection etc. It also implements MenuFactory. Menus use the ControlsAdapter interfaces to group together lower level commands to make them easy to use from the UI code. Menus are basically the viewmodels for the MenuFragments. I got a bit lazy about creating menu classes while doing the selection menus and they just use the ControlsAdapter interfaces directly, I might separate out the logic from the UI code in those ones later on.

There is a ControlsResolver which is used by fragments and activities to get access to the ControlsAdapter interfaces they need. Might be nice to use Dagger for this kind of thing eventually.


andreas-eberle:
@Sylapse: Is this PR code containing the one of #532?

andreas-eberle:
Would it be OK for you if I merge master into this branch?

Sylapse:
Didn't mean to do a second PR, just wasn't paying attention to the branch i was on!

It might be worth waiting before merging. I will push more changes tonight or tomorrow. I've got it able to save and load single player games and create and join network games now. Just needs a couple of loose ends sorting.

andreas-eberle:
OK, so I close this PR and you can push your changes to the branch of #532. Just tell me when I should try merging master into your branch. 

And thanks for all your work!

PR #545, opened 2017-02-14
Feature implemtation of stock building

andreas-eberle:


andreas-eberle:
@michaelzangl: The stock isn't rendered correctly when there is some stuff stored in it. How can we fix this?

PR #557, opened 2017-03-28
unified Movable interface

homoroselaps:
With this PR the architecture supports upcoming refactoring of the movable biotope.
Right now I'm working on a rewrite of the movable class. To support gradual transition to the new architecture the old Movable implementation needs a unified interface.


homoroselaps:
On my local setup all tests are passing. 
I can't find any meaningful error message in the travis output.
How should I proceed?

andreas-eberle:
@homoroselaps: I'll have a look. That is a bit weird.

jsettlers.logic/src/main/java/jsettlers/logic/movable/Movable.java
```diff
@@ -53,11 +54,10 @@
  * @author Andreas Eberle
  *
  */
-public final class Movable implements IScheduledTimerable, IPathCalculatable, IDebugable, Serializable, IViewDistancable, IGuiMovable,
-		IAttackableMovable {
+public final class Movable implements ILogicMovable {
 	private static final long serialVersionUID = 2472076796407425256L;
-	private static final HashMap<Integer, Movable> movablesByID = new HashMap<Integer, Movable>();
-	private static final ConcurrentLinkedQueue<Movable> allMovables = new ConcurrentLinkedQueue<Movable>();
+	private static final HashMap<Integer, ILogicMovable> movablesByID = new HashMap<Integer, ILogicMovable>();
```
andreas-eberle: I think we should stick to Movable here, as this is an internal map (same goes for allMovables). Only Movables can be stored here. What do you think?

andreas-eberle:
@homoroselaps: I fixed the AutoReplayIT files by just generating them again... maybe you generated them on a wrong branch or something like that...

homoroselaps:
thank you so much for your effort!

jsettlers.logic/src/main/java/jsettlers/logic/movable/Movable.java
homoroselaps: As the name suggests these fields contain all movables which are not exclusively used by the movable class. Other actors access them with `Movable.getMovableByID()`. The movables with the new architecture under the hood should be present in these fields, too. They are not a subclass of Movable, therefore `ILogicMovable` is necessary here. 
Later I decided to move the static fields into a separate `MovableDataManager` class in my commit b1f911c5c4b6c7aca27c22d4365e29fd5b49d83d

andreas-eberle: Ok, I understand.

PR #558, opened 2017-04-06
Make audio position dependend and fix some sounds

rpolzer:
object sound improved: carpenter, falling tree, etc
region sound improved: mountain, river, desert, etc.
position dependent sound
zoom dependent sound

### DIFF
--- a/go.graphics.swing/src/main/java/go/graphics/swing/sound/SwingSoundPlayer.java
+++ b/go.graphics.swing/src/main/java/go/graphics/swing/sound/SwingSoundPlayer.java
@@ -14,11 +14,6 @@
  *******************************************************************************/
 package go.graphics.swing.sound;
 
-import go.graphics.sound.ForgettingQueue;
-import go.graphics.sound.ForgettingQueue.Sound;
-import go.graphics.sound.ISoundDataRetriever;
-import go.graphics.sound.SoundPlayer;
-
 import java.io.IOException;
 
 import javax.sound.sampled.AudioFormat;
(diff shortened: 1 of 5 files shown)


andreas-eberle:

go.graphics.swing/src/main/java/go/graphics/swing/sound/SwingSoundPlayer.java
```diff
@@ -28,9 +23,14 @@
...
 
+import go.graphics.sound.ForgettingQueue;
+import go.graphics.sound.ForgettingQueue.Sound;
+import go.graphics.sound.ISoundDataRetriever;
+import go.graphics.sound.SoundPlayer;
+
 public class SwingSoundPlayer implements SoundPlayer {
 	private static final int BUFFER_SIZE = 4048 * 4;
-	private static final int SOUND_THREADS = 3;
+	private static final int SOUND_THREADS = 30;
```
michaelzangl: This won't work well on most phones.

jsettlers.graphics/src/main/java/jsettlers/graphics/sound/SoundManager.java
```diff
@@ -213,21 +233,84 @@ private static File getSoundFile() {
...
+	 * @param soundid
+	 *            The sound id to play
+	 * @param volume
+	 *            The volume
+	 * @param x
+	 *            The x coordinate of the sound
+	 * @param y
+	 *            The y coordinate of the sound
+	 */
+	public void playSound(int soundid, float volume, int x, int y) {
```
michaelzangl: Nice. This was missing a long time ;-)

go.graphics.swing/src/main/java/go/graphics/swing/sound/SwingSoundPlayer.java
rpolzer: 3 is definitely not enough, 10 might be a compromise and 30 sounds good - it is up to you what you select here - just make a sound check on a bit more crowded map

rpolzer: I thought this thread number would be only for PCs and not for phones, because it is within "swing" and not within "android"

andreas-eberle: @michaelzangl: I think @rpolzer is right about that this code will only be used by the Swing code, isn't he?

michaelzangl: Yes, he is. My bad ;-). You can keep it at 30 - current PCs should be able to handle so many sound channels.

andreas-eberle:
@rpolzer: Thanks for these changes!

PR #563, opened 2017-04-29
Movable Refactoring

homoroselaps:
This refactoring includes:
- a new architecture for game objects (Entity Component Model - similar to Unity's)
- a new way to define behavior (behavior tree - similar to [NPBehave](github.com/meniku/NPBehave))

The new architecture is meant to replace implementations of all movables.
For code exploration of the ECM you might use `EntityFactory.CreateMovable()` as a starting point.
For code exploration of the behavior tree you might use `DonkeyBehaviorComponent`.


jsettlers.logic/src/main/java/jsettlers/logic/movable/components/MultiMaterialComponent.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import jsettlers.common.material.EMaterialType;
+
+/**
+ * @author homoroselaps
+ */
+
+public class MultiMaterialComponent extends MaterialComponent {
```
ArneMayer, line 13: Does it make sense to refactor MaterialComponent and MultiMaterialComponent together to a single class?

jsettlers.logic/src/main/java/jsettlers/logic/movable/BehaviorTreeFactory.java
```diff
@@ -0,0 +1,116 @@
...
+        protected void onOpen(Tick<Entity> tick) {
+            done = false;
+        }
+    }
+
+    protected static Condition<Entity> TriggerCondition(Class<? extends Notification> type) {
+        return new Condition<>(entity -> entity.getNotificationsIt(type).hasNext());
+    }
+
+    protected static Debug $(String message, Node<Entity> child) {
```
ArneMayer: The name of the method does not express its behaviour.

jsettlers.logic/src/main/java/jsettlers/logic/movable/components/MultiMaterialComponent.java
homoroselaps: The class `MaterialComponent` implements the function `GetMaterial` of the `ILogicMovable` interface which is used mainly for choosing the correct visuals and for Strategies (like bearer and buildingworker) assuming, that a movable carries only one material at a time. For consistency reasons with the old implementation, I chose `MultiMaterialComponent` to be an extension of the base assumption of "only one material per movable". (although `GetMaterial` becomes ambiguous for more than one material carried)

jsettlers.logic/src/main/java/jsettlers/logic/movable/BehaviorTreeFactory.java
homoroselaps: Thank you for your annotation. You are completely right, this indicates bad code style.
In this decision I prioritized readability of the whole behavior tree over comprehensibility of this single function. Some behavior tree frameworks ship with a graphical tool, which we don't have. Therefore, I did my best to avoid as much boilerplate code/code repetition as possible, to support behavior design in code.
The function $ is here defined as a shorthand for the creation of a ´Debug´ node, which annotates tree nodes with debug information/comments.
I choose $ as to imitate the jQuery function, which developers might already be familiar with.
I'd fine with any other valid symbol, but would rather not clatter the tree with to many 
```new Debug(...)``` statements.

**Comment by rpolzer** [comment]
movables are getting permanently blocked when their destination is unreachable

File: jsettlers.logic/src/main/java/jsettlers/logic/movable/Entity.java [hunk 3]
```diff
@@ -0,0 +1,274 @@
...
+
+    public final BearerComponent bearerC() {
+        return get(BearerComponent.class);
+    }
+
+    public final SpecialistComponent specC() {
+        return get(SpecialistComponent.class);
+    }
+
+    public final GameFieldComponent gameC() {
```
andreas-repp: Please give these methods full names. E.g. getGameFieldComponent()

jsettlers.logic/src/main/java/jsettlers/logic/movable/EntityFactory.java
```diff
@@ -0,0 +1,97 @@
...
+import jsettlers.logic.player.Player;
+
+/**
+ * @author homoroselaps
+ */
+
+public final class EntityFactory {
+    private EntityFactory() {}
+
+    public static ILogicMovable CreateMovable(AbstractMovableGrid grid, EMovableType movableType, ShortPoint2D position, Player player) {
```
andreas-repp: In Java method names start with a lower case letter

jsettlers.logic/src/main/java/jsettlers/logic/movable/BehaviorTreeHelper.java
```diff
@@ -0,0 +1,239 @@
...
+        public void onOpen(Tick<Context> tick) {
+            endTime = MatchConstants.clock().getTime() + delay;
+        }
+    }
+
+    public static Node<Context> WaitForNotification(Class<? extends Notification> type, boolean consume) {
+        return Wait(NotificationCondition(type, consume));
+    }
+
+    public static Debug $(String msg, Node<Context> child) {
```
andreas-repp: please name this method debug or something like that. I first thought this is a crazy new syntax...

andreas-eberle:
Please format your code with our autoformatter (e.g. tabs instead of spaces and so on).

jsettlers.logic/src/main/java/jsettlers/logic/movable/simplebehaviortree/IBooleanConditionFunction.java
```diff
@@ -0,0 +1,7 @@
+package jsettlers.logic.movable.simplebehaviortree;
+
+import java.io.Serializable;
+import java.util.function.Function;
```
andreas-repp: We have to use the **java8**.* package to be able to run the code on Android (the java8 package comes from the streamsupport library and offers a backport of the Java8 possibilities). The same applies for the @FunctionalInterface. We need to explicitly import it with `import java8.lang.FunctionalInterface;`

jsettlers.logic/src/main/java/jsettlers/logic/movable/simplebehaviortree/Tick.java
```diff
@@ -0,0 +1,52 @@
+package jsettlers.logic.movable.simplebehaviortree;
+import java.io.Serializable;
+import java.util.LinkedList;
+import java.util.Stack;
+
+public class Tick<T> implements Serializable {
+	private static final long serialVersionUID = 3673558738736795584L;
+
+	public final Root<T> Root;
```
andreas-repp: fields and variables also start with a lower case letter.

jsettlers.logic/src/main/java/jsettlers/logic/movable/simplebehaviortree/IBooleanConditionFunction.java
```diff
@@ -0,0 +1,7 @@
...
+
+@FunctionalInterface
+public interface IBooleanConditionFunction<T> extends Function<T,Boolean>, Serializable { }
```
andreas-repp: Extend Predicate<T> instead of Function<T,Boolean> and give it a similar name

jsettlers.logic/src/main/java/jsettlers/logic/movable/BehaviorTreeHelper.java
```diff
@@ -0,0 +1,239 @@
...
+            c.entity.aniC().startAnimation(animation, duration);
+        });
+    }
+
+    public static void convertTo(Entity entity, EMovableType type) {
+        Entity blueprint = EntityFactory.CreateEntity(entity.gameC().getMovableGrid(), type, entity.movC().getPos(), entity.movC().getPlayer());
+        entity.convertTo(blueprint);
+    }
+
+    public static <T> Node<T> Optional(Node<T> child) {
```
andreas-repp: This should have a different name. An Optional is a fixed thing in java8, which makes this confusing.

jsettlers.logic/src/main/java/jsettlers/logic/movable/simplebehaviortree/nodes/MemSelector.java
```diff
@@ -0,0 +1,30 @@
...
+	@SafeVarargs
+	public MemSelector(Node<T>... children) {
+		super(children);
+	}
+	
+	@Override
+	protected NodeStatus onTick(Tick<T> tick) { 
+		for (; index < children.size(); index++) {
+			NodeStatus status = children.get(index).execute(tick);
+			if (!status.equals(NodeStatus.Failure))
```
andreas-repp: Please always add curly brackets after ifs.

jsettlers.logic/src/main/java/jsettlers/logic/movable/components/Component.java
```diff
@@ -0,0 +1,119 @@
...
+        return result;
+    }
+
+    public <T extends Notification> T getNextNotification(Class<T> type, boolean consume) {
+        Iterator<T> it = getNotificationsIt(type);
+        T note = it.next();
+        if (note != null) {
+            if (consume) consumeNotification(note);
+        }
+        return null;
```
andreas-repp: You never return a notification... even if you find one.

jsettlers.logic/src/main/java/jsettlers/logic/movable/components/Component.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    public <T extends Notification> boolean containsNotification(Class<T> type) {
+        return getNotificationsIt(type).hasNext();
+    }
+
+    public <T extends Notification> Iterator<T> getNotificationsIt(Class<T> type) {
```
andreas-repp: Some things here:

* I think it would be better to realize this with a java8 stream. From what I found, you always only check if there is a single element and then get that element. You could make this function return a Stream<T> and build the stream like this:

```
   return stream(entity.getAllNotifications()).filter(type::isInstance).map(notification -> (T)notification);
```

and when you want to get the first matching notification just do

```
getNotificationsIt(type).findAny().isPresent()
```

Furthermore, I would then rename this method to something like getNotificationsOfType.

* Regarding the consumed status of notifcations. Have you thought about adding a consumed flag to the Notification parent class along with a `consume()` method? This way you can save this state of the notification in the notification, where it belongs (more OOP). (and it should be better in performance compared to the hash set lookup).

* Is there a need to use `type.isInstance()`? I mean will there be the use case that you ask for a supertype notification and expect to get a child class of it? If not, we could simply do a == comparision on the class of the notification, which should be considerably faster.

jsettlers.logic/src/main/java/jsettlers/logic/movable/MovableWrapper.java
```diff
@@ -0,0 +1,256 @@
...
+        ois.defaultReadObject();
+        entity.get(GameFieldComponent.class).getMovableMap().put(entity.getID(), this);
+        entity.get(GameFieldComponent.class).getAllMovables().add(this);
+    }
+
+    //region Interface Implementations
+    @Override
+    public short getViewDistance() {
+        if (!entity.containsComponent(MovableComponent.class)) return 0;
+        return entity.get(MovableComponent.class).getViewDistance();
```
andreas-repp: I would change these methods to directly do the get on entity and then check if the result is null. If so, return the default value, if it isn't null, call the method. This way, the code does not need to query the set twice.

jsettlers.logic/src/main/java/jsettlers/logic/movable/components/BearerBehaviorComponent.java
```diff
@@ -0,0 +1,239 @@
...
+                    )),
+                    $("handle failure", Sequence(
+                        Action(BearerBehaviorComponent::resetJob)
+                    ))
+                )
+            )
+        ));
+    }
+
+    private static Action<Context> accept_SaveDeliveryJob() {
```
andreas-repp: Please don't use underscores in Java method names. Camel case is the way to go. Furthermore, accept_Save seems kind of doubled.

andreas-eberle:
@homoroselaps: I really like the idea of using behavior trees! It really offers a way to do more complex stuff in a much cleaner way. However, I was wondering, isn't their a working library out there already that we could use? I think if such a library does not have major drawbacks, it's probably better to use it off the shelf instead of reimplementing it. Do you know about such libaries?

andreas-eberle:
@homoroselaps: It would be great, if you could reach out to me on Discord so I can ask you some questions about your code.

jsettlers.logic/src/main/java/jsettlers/logic/movable/components/Component.java
homoroselaps: > Regarding the consumed status of notifcations.

It is important that each component is able to track which notifications got consumed so far individually. Right now the `GetNotificationsOfType` returns the objects from the entity. This ensures the equality of `Notifications` across `Component` borders. If the consumed flag is part of the `Notification` itself, then we have to override the standard equality of Notifications.

>Is there a need to use type.isInstance()? I mean will there be the use case that you ask for a supertype notification and expect to get a child class of it?

Yes this is the intended behavior. This way eg a `Notification` GoToPostion::GoToPositionWithMode can be treated by the `DefaultSteeringComponent` simply as `GoToPosition` as well as by an `SpecialSteeringComponent` which supports different modes of steering.

PR #564, opened 2017-05-06
Swing: Zoom centered at mouse pointer

rpolzer:
up to now, zooming was centered at the screen center - now the zoom center is where the mouse points to (swing only)


go.graphics.android/src/main/java/go/graphics/android/GOSurfaceView.java
```diff
@@ -191,7 +191,7 @@ public void onTouchEvent(MotionEvent e) {
 					} else if (e.getAction() == MotionEvent.ACTION_UP) {
 						endPan(computePanPoint(e));
 						if (doZoom) {
-							endZoomEvent(factor);
+							endZoomEvent(factor, null);
```
andreas-eberle: @rpolzer: Is it correct to use null here?

andreas-eberle:
@rpolzer: Thanks for the PR. The change feels good in the game.

@michaelzangl: Can you have a look at the code? You know these parts better than I do.

go.graphics.android/src/main/java/go/graphics/android/GOSurfaceView.java
rpolzer: null is intended in any cases where zoomEvent is called without a known mouse or pointer position

PR #566, opened 2017-05-10
toolsmith produces not only for workers but also UI requested tools

rpolzer:
If workers need a tool, the toolsmith pruduces it as before. But if tool production is requested by UI settings, these tools are now also produced, but with lower priority.


jsettlers.logic/src/main/java/jsettlers/logic/buildings/MaterialProductionSettings.java
```diff
@@ -63,10 +72,16 @@ public float resultingRatioOfMaterial(EMaterialType type) {
 				return 0;
 			}
 			return (configuredRatioOfMaterial(type) / allWeapons);
+		} else {
```
andreas-eberle: We should check if the given type is in the TOOLS_LIST.

jsettlers.logic/src/main/java/jsettlers/logic/buildings/MaterialProductionSettings.java
```diff
@@ -125,6 +138,8 @@ private EMaterialType getMaterialOfGroupToProduce(EMaterialType[] materialGroup)
 		for (float ratio : materialRatio) {
 			maxRatio += ratio;
 		}
+		if (maxRatio == 0)
```
andreas-eberle: Please add the { } around the statement

andreas-eberle:
I replaced the arrays and lists of EMaterialType with the usage of EnumSets and moved the EnumSets to EMaterialType.

andreas-eberle:
In order to fix the AutoReplayIT, you can regenerate the reference files by executing the main in jsettlers.integration.replay.RegenerateAutoReplayITReferences in the jsettlers.tools project. Keep in mind to only do this if you expect the behavior/the generated savegame to change.

andreas-eberle:
At the moment the code is that the user's setting is only used when the automated tool production does not require a certain tool. Shouldn't we do it the other way around? If I set the tool production manually, I'd expect the toolsmith to produce exactly as I said. What do you think @rpolzer?

rpolzer:
I would prefer to have UI ordered tools produced first - the only reason I did it the other way round was that I suspected this could block the game.

andreas-eberle:
Why do you think so? I mean, if the user sets a ratio for a tool the automated tool production will not be used any more. But in that case the user wants the game to produce the set ratio. And if the user doesn't want that any more he can set the ratio back to zero and the automated production will take over again.

rpolzer:
I think it would be best if both works at the same time: user requested and worker requested production. So if the user has set a ratio and a worker needs a tool, this tool for the worker should be produced. What about giving user requested tools set as numbers first priority, then worker requested tools with medium priority and handle a user requested production ratio with lowest priority?

andreas-eberle:
That sounds like a good idea. Can you implement it?

rpolzer:
this should be as requested

andreas-eberle:
@rpolzer: When the AutoReplayIT reference files are in conflict, you have to generate them again with the merged code. This is because then both branches have changes to the replay references and thus none of them are valid with the other changes.

andreas-eberle:
@rpolzer: I refactored the MaterialProductionSettings class as it was a bit messy. Can you check if my implementation still works as we discussed? (Just to make sure I didn't miss something.)

rpolzer:
I checked the toolsmith work and gave the waeponsmith an update. Both look good now.

Issue #310, opened 2016-03-11
GUI disappears on game start


iandioch:
I tried to start the game, using both assets from the Amazons demo, and also from a physical CD, and both times the following happened:
- I started the JAR
- The menu showed ok, with the button assets showing fine
- I started a new single player game with several different maps, it didn't seem to make a difference
- When the game loaded, it would display fine for a split second
- The left menu would disappear, and all of the entities on the screen would disappear, leaving just the map
- The game would still run - I could hear the sound effects, and the people still left footprints. Image [here](http://i.imgur.com/99p0CjM.jpg)

andreas-eberle:
Hi @iandioch,

can you send me the logfiles? They are in the folder `resources/logs`. (andreas.eberle [at] settlers-android-clone.com)

Best regards

iandioch:
Sorry for the delayed response. I didn't know the logs there existed. The relevant part of the log is below.

```
Could not find/load graphic file 13
Could not find/load graphic file 13
Could not find/load graphic file 13
Could not find/load graphic file 13
Could not find/load graphic file 13
Could not find/load graphic file 13
Could not find/load graphic file 13
Could not find/load graphic file 13
Could not find/load graphic file 13
Could not find/load graphic file 13
Could not find/load graphic file 13
Could not find/load graphic file 13
Main draw handler cought throwable:
java.lang.ArithmeticException: / by zero
	at jsettlers.graphics.map.draw.MapObjectDrawer.drawPlayerableWaving(MapObjectDrawer.java:1042)
	at jsettlers.graphics.map.draw.MapObjectDrawer.drawMapObject(MapObjectDrawer.java:298)
	at jsettlers.graphics.map.MapContent.drawTile(MapContent.java:433)
	at jsettlers.graphics.map.MapContent.drawMain(MapContent.java:403)
	at jsettlers.graphics.map.MapContent.drawContent(MapContent.java:220)
	at go.graphics.region.Region.drawRegion(Region.java:109)
	at go.graphics.area.Area.drawRegionAt(Area.java:133)
	at go.graphics.area.Area.drawArea(Area.java:121)
	at go.graphics.swing.AreaContainer.draw(AreaContainer.java:163)
	at go.graphics.swing.AreaContainer$1.display(AreaContainer.java:96)
	at jogamp.opengl.GLDrawableHelper.displayImpl(GLDrawableHelper.java:691)
[...]

PR #567, opened 2017-05-11
Make search for graphics and sound folders case insensitive and unify it between Android and desktop version (fixes #489)

### LINKED ISSUE #489: Support upper/lower case file names for graphics.
Under Linux several things are not rendered (hidden?):
broken settlers rendering
This is the corresponding log:

```
$ java -jar JSettlers.jar 
No default command line options file found.
Options loaded: {}
Source: (file:/***/JSettlers.jar <no signer certificates>)
Could not find/load graphic file 01
Could not find/load graphic file 11
java.lang.NullPointerException
    at jsettlers.graphics.map.draw.ImagePreloadTask.run(ImagePreloadTask.java:121)
    at java.lang.Thread.run(Thread.java:745)
Could not find/load graphic file 12
java.lang.NullPointerException
    at jsettlers.graphics.map.draw.ImagePreloadTask.run(ImagePreloadTask.java:189)
    at java.lang.Thread.run(Thread.java:745)
Could not find/load graphic file 03
Could not find/load graphic file 03
Warning: Could not load LocaleSuffix [locale=en_US, useCountry=true]. Falling back to next file.
Loading Sequence number 17
Loading Sequence number 18
```

Under Windows, with the same data files, it works fine.

rpolzer:
resolves issue #489: gfx files may have arbitrary upper or lower case file names


rpolzer:
seems to resolve issue #310, too

andreas-eberle:
@rpolzer: Many thanks for all your PRs! I'll review them as soon as possible! But I first want to finish the #532. If you have any questions or so, feel free to join our discord https://discord.gg/2hVV4u6.

jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/ImageProvider.java
```diff
@@ -300,9 +300,10 @@ public void invalidateAll() {
 
 	private File findFileInPaths(String fileName) {
 		for (File path : this.lookupPaths) {
-			File searched = new File(path, fileName);
-			if (searched.isFile() && searched.canRead()) {
-				return searched;
+			for (File f : path.listFiles()) {
+				if (f.getName().equalsIgnoreCase(fileName) && f.isFile() && f.canRead()) {
```
andreas-eberle: The condition of this should better be
`f.isFile() && f.canRead() && fileName.equalsIgnoreCase(f.getName()) `
This way, we can prevent a potential NullPointer if f.getName returns null and the simpler checks are done first.

jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/ImageProvider.java
```diff
@@ -300,9 +300,10 @@ public void invalidateAll() {
 
 	private File findFileInPaths(String fileName) {
 		for (File path : this.lookupPaths) {
-			File searched = new File(path, fileName);
-			if (searched.isFile() && searched.canRead()) {
-				return searched;
+			for (File f : path.listFiles()) {
```
andreas-eberle: Please don't use single letter variable names for this. Calling it currentFile is more obvious (I know, here isn't much code around, but we want to keep a clean code in the project ;) )

jsettlers.main.swing/src/main/java/jsettlers/main/swing/resources/SettlersFolderChecker.java
```diff
@@ -64,13 +62,16 @@ public static SettlersFolderInfo checkSettlersFolder(String settlersBaseFolder)
 
 		for (File f : files) {
 			if (f.getName().equalsIgnoreCase("snd")) {
-				File testSndFile = new File(f, "Siedler3_00.dat");
-				if (testSndFile.exists()) {
-					sndFolder = f;
+				for (File g : f.listFiles()) {
```
andreas-eberle: Same as before, please no one letter variable names (except for index i in for loop)

jsettlers.main.swing/src/main/java/jsettlers/main/swing/resources/SettlersFolderChecker.java
```diff
@@ -64,13 +62,16 @@ public static SettlersFolderInfo checkSettlersFolder(String settlersBaseFolder)
...
+					g.getName();
```
andreas-eberle: Did you mean to have this line here?

jsettlers.main.swing/src/main/java/jsettlers/main/swing/resources/SettlersFolderChecker.java
```diff
@@ -64,13 +62,16 @@ public static SettlersFolderInfo checkSettlersFolder(String settlersBaseFolder)
...
+					if (g.getName().equalsIgnoreCase("siedler3_00.dat")) {
```
andreas-eberle: Again, switch the argument and the object to 
`"siedler3_00.dat".equalsIgnoreCase(g.getName())`

jsettlers.main.swing/src/main/java/jsettlers/main/swing/resources/SettlersFolderChecker.java
```diff
@@ -64,13 +62,16 @@ public static SettlersFolderInfo checkSettlersFolder(String settlersBaseFolder)
...
+						sndFolder = f;
+					}
 				}
 			} else if (f.getName().equalsIgnoreCase("gfx")) {
-				for (DatFileType t : DatFileType.values()) {
-					if (new File(f, "siedler3_00" + t.getFileSuffix()).exists()) {
+				for (File g : f.listFiles()) {
+					String a = g.getName();
+					if (a.substring(0, a.indexOf('.')).equalsIgnoreCase("siedler3_00")) {
```
andreas-eberle: Same as above.

rpolzer:
have a look...

andreas-eberle:
The changes look good. One more thing however: While reviewing one of the other PRs, I saw that the sound file is also searched in SoundManager:221. So the case insensitve check would also be needed there. However, I think it would be best to consolidate these codes into the SettlersFolderChecker class to have it all in one place. Can you do this, @rpolzer?

rpolzer:
I tried this, but the result is either a gradle error
`Error:(31, 38) error: package jsettlers.main.swing.resources does not exist`
`Error:(192, 18) error: cannot find symbol variable SettlersFolderChecker`
or a circular dependency
`Error:Circular reference between projects: :jsettlers.main.swing -> :jsettlers.graphics -> :jsettlers.main.swing`
because the module graphics (which needs the file lookup) can not depend on the module main.swing (where I tried to put the code).

So the function
`File findFileInPaths(String fileName, List<File> lookupPaths)`
should reside somewhere in the module graphics and not in the class SettlersFolderChecker.

andreas-eberle:
I'll try to sort this out.

andreas-eberle:
@rpolzer: I had to rework and sort out a lot of old messy code... However, now I think it should be ok and the same code is running on the desktop and the Android version. Can you have a look again to double-check?

rpolzer:
It is still not possible to have findFileInPath inside the class SettlersFolderChecker.
In my opinion, this is not necessary and I would leave it where it is.
I tested this branch on Linux: changing of graphics file names upper and lower case letters is OK.

Issue #446, opened 2016-05-21
Sounds give away the geography of the map


derfuhrer2007:
Sounds seem to give away the geography of an unexplored part of the map (still black).

You can be over the sea and hear seagulls and so on.

andreas-eberle:
Was implemented in #574.

PR #568, opened 2017-05-11
resolve issue #446

rpolzer:
invisible movables are silent now


andreas-eberle:
@rpolzer: Thanks for this fix. I changed it slightly, so that sounds are only played when the area is currently visible (the value must be higher than just explored). This is to prevent that you're able to hear the settllers of an enemy just because you once saw an area.

PR #569, opened 2017-05-11
Settler Statistic Panel

nptr:
Added the ingame settler statistic panel for the romans, and the layout for the amazons.
The only information missing is the amount of "beds" for the civilians as i think this concept is not
implemented in the game yet.


andreas-eberle:
@nptr: Thanks for this PR. Looks good to me. I'm looking forward to your next PRs ;)

Issue #565, opened 2017-05-06
Wrong images shown for soldiers occupying a tower


ziarrek:
As seen in a small tower:
![tower_occupying_images](https://cloud.githubusercontent.com/assets/4028673/25773051/b5e58852-326e-11e7-9437-22bf67ddaa88.png)

The tower menu shows wrong images for the types of soldiers occupying the tower.
Tested (map 2Hours, Swing UI built and running on Ubuntu):
- level 1 swordsman shows as level 1 archer
- level 1 archer shows as level 2 pikeman
- level 3 swordsman shows as level 3 archer

andreas-eberle:
I'm closing this in favor of #573 which targets to fix the underlying problem. @ziarrek: Thanks for reporting.

PR #571, opened 2017-05-13
resolve issue #565

rpolzer:
resolve issue #565: soldiers on marked tower show the right images


andreas-eberle:
@rpolzer: Unfortunately this doesn't fix the real issue. The problem is that there are several version of The Settlers III and each of them has slightly different indexes in their graphics files. So the new indexes you entered only work in the version you have and don't work for example in the Amazon Demo. What we really need is a system to detect the type of files we have available and then choose the appropriate set of indexes. I've opened issue #573 to keep track of this. Feel free to implement a fix for that ;)

rpolzer:
should we close this issue?

PR #572, opened 2017-05-14
waterworker uses empty bucket

rpolzer:


andreas-eberle:
@rpolzer: Thanks, looks fine. I also adapted the animation duration of the waterworker from 4 to 1 second and fixed the tests.

PR #574, opened 2017-05-17
play sound when visible

rpolzer:
change ">" and "<" in playSound


andreas-eberle:
Thanks for fixing my mistake. Can you make it a <=, so no sound is played when the area is only explored?

PR #588, opened 2017-05-29
Make it easier to place buildings on mobile

### LINKED ISSUE #102: Construction state is left even if the building could not be placed on the target position
Steps to reproduce:
- Select a building to construct in the construction menu.
- Tap on a location, where this building cannot be build.

Now the construction state is left, but no building is constructed. This makes it necessary to selected the building in the menu again.

Furthermore: in the Swing UI, we do not leave the construction state, when a building has been placed. Shouldn't it be the same on Android?

andreas-eberle:
fixes #102.

Changed the way building placement is done to search an area around the selected position for a valid placement position. This area is also larger on mobile to compensate that players don't have a mouse there.

### DIFF
--- a/jsettlers.common/src/main/java/jsettlers/common/position/ShortPoint2D.java
+++ b/jsettlers.common/src/main/java/jsettlers/common/position/ShortPoint2D.java
@@ -125,6 +125,10 @@ public class ShortPoint2D implements Serializable {
 		}
 	}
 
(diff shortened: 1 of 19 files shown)


andreas-eberle:
@Tom-Pratt: Can you have a look at this?

Tom-Pratt:
Works nice 👍

Issue #548, opened 2017-02-22
Support high dpi displays


andreas-eberle:
On my 15" display with a 4k resolution, the texts of the game are very small.

PR #600, opened 2017-06-08
Added scaling factor calculation and font size scaling.

PayDay1:
Not sure if the initial calculation is at the right place.
Should improve #548.

### DIFF
--- a/go.graphics.swing/src/main/java/go/graphics/swing/text/JOGLTextDrawer.java
+++ b/go.graphics.swing/src/main/java/go/graphics/swing/text/JOGLTextDrawer.java
@@ -33,6 +33,7 @@ import java.util.ArrayList;
  */
 public final class JOGLTextDrawer implements TextDrawer {
 
+<<<<<<< HEAD
 	private static final int INTERVAL_SCALING_UPDATE = 10000;
 	private static final int DEFAULT_DPI = 96;
 	private static final String FONTNAME = "Arial";
@@ -52,6 +53,17 @@ public final class JOGLTextDrawer implements TextDrawer {
 				}
 			}
 		}, 0, INTERVAL_SCALING_UPDATE);
+=======
+	private static final String FONT_NAME = "Arial";
+
+	private static final int DEFAULT_DPI_WIDTH = 1920;
+	private static final float SCALING_FACTOR;
+
+	static {
+		GraphicsDevice graphicsDevice = GraphicsEnvironment.getLocalGraphicsEnvironment().getDefaultScreenDevice();
+		int screenWidth = graphicsDevice.getDisplayMode().getWidth();
+		SCALING_FACTOR = Math.min((float) screenWidth / DEFAULT_DPI_WIDTH, 1);
+>>>>>>> 0381dce4d157ce2e7caf66db40ff9fbb1cb7ecfc
 	}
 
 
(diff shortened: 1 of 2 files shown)


go.graphics.swing/src/main/java/go/graphics/swing/text/JOGLTextDrawer.java
```diff
@@ -19,32 +19,43 @@
...
  * @author michael
  */
 public final class JOGLTextDrawer implements TextDrawer {
 
 	private static final String FONTNAME = "Arial";
 
+	private static final int DEFAULT_DPI_WIDTH = 1920;
+	private static float SCALING_FACTOR = 1;
+
+	public static void calculateScalingFactor() {
```
andreas-eberle: should we just use a static{} block here. Then this will be calculated when the class if first used. I don't like having to call this method from SwingManagedJSettlers.

andreas-eberle:
@PayDay1: Cool, this works great.

go.graphics.swing/src/main/java/go/graphics/swing/text/JOGLTextDrawer.java
```diff
@@ -19,32 +19,43 @@
 import go.graphics.text.EFontSize;
 import go.graphics.text.TextDrawer;
 
-import java.awt.Font;
+import java.awt.*;
```
andreas-eberle: Please configure the auto formatter as described in https://github.com/jsettlers/settlers-remake/wiki/Developer's-Guide#development-guidelines

go.graphics.swing/src/main/java/go/graphics/swing/text/JOGLTextDrawer.java
```diff
@@ -20,31 +20,43 @@
...
 
-	private static final String FONTNAME = "Arial";
+	private static final String FONT_NAME = "Arial";
+
+	private static final int DEFAULT_DPI_WIDTH = 1920;
+	private static final float SCALING_FACTOR;
+
+	static {
+		GraphicsDevice graphicsDevice = GraphicsEnvironment.getLocalGraphicsEnvironment().getDefaultScreenDevice();
+		int screenWidth = graphicsDevice.getDisplayMode().getWidth();
```
michaelzangl: Are you sure computing this in advance is a good idea? I think it should depend on the main viewport size and be re-computed after the window size is changed.

andreas-eberle: The scale doesn't depend on the window size. So we don't need to refresh when the JSettlers window changes its size. However, you're right that we should recalculate it when the used screen changes. Do you know how if we can get an event when the user moves the window to another screen?

andreas-eberle: From what I found on the internet, the best idea is to regularily check / recalculate the scale. Another thing: @PayDay1: I just found `Toolkit.getDefaultToolkit().getScreenResolution();` which returns a dpi number. I think using this would be better than using the screen resolution (a screen can have 4k but be 50 inch large)...

PayDay1: That's even better, yeah. We can then just scale it with 96 as the base dpi.

andreas-eberle:
@PayDay1: Any progress on this?

andreas-eberle:
@PayDay1: What is the status on this? Can you find the time to get it finished, so we can merge it?

andreas-eberle:
@PayDay1: I'd really like to get this in. I myself have these problems with my display ;)

andreas-eberle:
@PayDay1: Do you have plans for a new PR on this topic?

PayDay1:
There is a new one open, although it isn't updating when the dpi changes, because re-creating the fonts didn't work quite yet.

PR #625, opened 2017-07-01
Stock building menu - android

tom-pratt:
I did this slightly different to the original controls. You just tap a material to toggle between stocked and not-stocked. In the original you select a material and then tap the stock/dont stock buttons.

Less clicks and more intuitive i think


andreas-eberle:
@tom-pratt: I like it that you only have to click once, that's great!

When checking your implementation, I found a problem: In the desktop version, we now have the possibility (in contrast to the original game) to set the stocked materials globally and locally for every stock. So in the desktop version you can go to the materials menu section and then to the stock section and configure the gobal stocking settings there. However, if you select a stock building and set the settings there, they are only local settings. 

At the moment you can only set the stock settings by selecting a stock and set it there. However, in contrast to the desktop version this sets the global settings. 

So I propose doing about the same as in the desktop version. Can you add the same stock menu as a page to the "goods" tab? If the user configures the stock settings there, the global settings should be changed (`local` flag of `SetAcceptedStockMaterialAction` must be false). If the user selects a stock and changes the settings there, they should be set locally (`local` flag true).

tom-pratt:
Ah yeah this makes sense. I saw the `local` bool when I was doing it and I thought that  was what it was for, but maybe it wasn't hooked up yet. Actually I had just forgotten that in the original game you control the stock from the goods menus and not by clicking on the stock building!

The stock menu will be easy to reuse as one of the goods menus as well

tom-pratt:
Ah now i remember (just had a look at desktop jsettlers). The stock is combined with the carrier priority order menu! The stock in and out buttons make more sense here as there are the priority actions which can also be performed.

I will change the stock building selection menu to be `local` true. Then I will do that other menu as a separate PR

andreas-eberle:
I don't think we should combine the priority menu and the stock menu on Android like it is done in the original game. I think we can split the functionality up into two menus. This makes it easier to design and use the UI and we don't have the need to stay at the same number of menus. Do you want to add the global menu looking just the same way as the local ones? Or should I merge this one first and you'll implement that next?

tom-pratt:
Yeah ok I like that  idea.

Merge this one anyway because its all working and I'm not sure how long till I'll next work on it.

PR #642, opened 2017-07-12
Global stock menu android

tom-pratt:
Working


jsettlers.main.android/src/main/java/jsettlers/main/android/gameplay/ui/adapters/MaterialsAdapter.java
```diff
@@ -0,0 +1,119 @@
+package jsettlers.main.android.gameplay.ui.adapters;
```
andreas-eberle: Please add the copyright notice.

jsettlers.main.android/src/main/java/jsettlers/main/android/gameplay/ui/viewholders/MaterialViewHolder.java
```diff
@@ -0,0 +1,38 @@
+package jsettlers.main.android.gameplay.ui.viewholders;
```
andreas-eberle: Copyright notice

jsettlers.main.android/src/main/java/jsettlers/main/android/gameplay/viewstates/StockMaterialState.java
```diff
@@ -0,0 +1,29 @@
+package jsettlers.main.android.gameplay.viewstates;
```
andreas-eberle: Copyright notice

andreas-eberle:
Two things:
* It should only be possible to set and see the storage configuration of the own player => when the center is over the land of of another player or no player, don't display the storage menu.
* The settings are actually only partition global and not map / game global. Do you think we should express that in the title? Or do you think everybody understands global to be per partition?

tom-pratt:
I'll make it so when you're outside one of your own partitions the menu hides and it shows a message saying something like "Storage settings must be set in one of your own settlements"

That addresses both issues without showing an overly long title.

michaelzangl:
_settlement_ -> I like that term for our technical _partitions_. We should make a list of terms that are used in the UI to make them consistent.

tom-pratt:
I think this is ready to merge

andreas-eberle:
I'll review it as soon as I find the time ;)

andreas-eberle:
@tom-pratt: Can you make the global stock menu the first tab in the materials menu? This should only be temporary, but at the moment the menu is all the way back as the last tab with all empty tabs before it. I think the tabs with functionallity should come first.

PR #648, opened 2017-07-22
Rework swing ui updating and fix labels of MaterialDistributionPanel

### LINKED ISSUE #521: Resources in newly claimed areas are not counted in statistics

### LINKED ISSUE #534: Wood cutters produced log count doesnt increase while selected
As long as the building selection content is open for the wood cutter, it's number of outputted logs doesnt increment. If you select something else then reselect the woodcutter is reflects the true count.

The same thing happened while working on Android version so probably not just a graphical code bug. Think the wrong data is coming through in the StackStates.

capture

### LINKED ISSUE #535: Material stack states dont get updated when FAST_FORWARD happens
If you select a building and fast forward before it receives any construction materials, the planks/stones counts never decrease even though they arrived during the fast forward.

Probably related to #534

### LINKED ISSUE #646: In game menus are not refreshing correctly with a fixed period.

andreas-eberle:
fixes #646 
fixes #521 
fixes #534 
fixes #535


andreas-eberle:
@michaelzangl: Can you have a look at the graphics changes?

jsettlers.common/src/main/java/jsettlers/common/position/IPositionSupplier.java
```diff
@@ -0,0 +1,23 @@
+/*
```
michaelzangl: We might want to decide on a common style here (`/*` or `/*********...`

andreas-eberle: /* is the new style (which IntelliJ uses with its generator). I use the new one for new files but won't change it for old files (if IntelliJ isn't doing it because I move a file).

PR #650, opened 2017-07-23
Android ships UI

tom-pratt:
not quite finished yet but just thought id create this PR so you can merge into yours when you want to.


tom-pratt:
cant get this right!

PR #657, opened 2017-07-25
Allow force-to moves for movables

michaelzangl:
Allow movables to be forced to move at a given position and pass on the flag to make them work there.

Logic (movables strategies) only implemented for Geologist.

### DIFF
--- a/jsettlers.graphics/src/main/java/jsettlers/graphics/action/MoveToAction.java
+++ b/jsettlers.graphics/src/main/java/jsettlers/graphics/action/MoveToAction.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2015
+ * Copyright (c) 2017
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"),
  * to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense,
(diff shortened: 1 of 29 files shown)


jsettlers.graphics/src/main/java/jsettlers/graphics/action/MoveToAction.java
```diff
@@ -0,0 +1,49 @@
...
+	public String toString() {
+		return "MoveToAction [mode=" + mode + ", getPosition()=" + getPosition() + "]";
+	}
+
+	public static EMoveToMode modeForModifiers(int modifiers) {
+		if ((modifiers & GOEvent.MODIFIER_CTRL) != 0) {
+			return EMoveToMode.WORK;
+		}
+		if ((modifiers & GOEvent.MODIFIER_SHIFT) != 0) {
+			return EMoveToMode.FORCED;
```
michaelzangl: Original: CTRL forces soldiers to go to that position

jsettlers.graphics/src/main/java/jsettlers/graphics/action/MoveToAction.java
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	@Override
+	public String toString() {
+		return "MoveToAction [mode=" + mode + ", getPosition()=" + getPosition() + "]";
+	}
+
+	public static EMoveToMode modeForModifiers(int modifiers) {
+		if ((modifiers & GOEvent.MODIFIER_CTRL) != 0) {
+			return EMoveToMode.WORK;
```
michaelzangl: Original: ALT

andreas-eberle:
Is this a duplicate of #709?

michaelzangl:
Yes, might be...

PaweX:
In the original Settlers III, we don't need to press ALT key, because "force move" feature is always on. Settlers IV introduced ALT+move as things are slightly different there.

PR #669, opened 2017-08-02
Added danish translation to Android GUI and mapeditor

stumpdk:
Note that I have only created the translation files.
If any configuration is necessary in order for the game and mapeditor to find the files, this has to be done.


stumpdk:
@andreas-eberle All language changes are now implemented in the same branch.

PR #671, opened 2017-08-04
Swing graphics

andreasb242:
Graphics from Ben


andreas-eberle:
Thanks @andreasb242 and Ben for these changes! Looks awesome!

One thing I noticed: When I try to load a savegame that cannot be loaded (e.g. I created the savegame with an older game version), I get the following error dialog:

grafik

Could you make the texts in the dialog localized? I don't think it needs to be skinned.

jsettlers.main.swing/src/main/java/jsettlers/main/swing/lookandfeel/ui/ButtonUiStone.java
```diff
@@ -15,138 +15,161 @@
...
+			UiImageLoader.get("sr_ui_button/sr_ui_button-border-left.png")
+	};
 
 	/**
-	 * Offset
+	 * Border images if the Button is pressed
 	 */
-	private int shiftOffset = 0;
+	private final BufferedImage[] BORDER_DOWN = { UiImageLoader.get("sr_ui_button_down/sr_ui_button-corner-upper-left.png"),
+			UiImageLoader.get("sr_ui_button_down/sr_ui_button-border-top.png"),
```
andreas-eberle: what does the sr_ in the folder names stand for?

andreas-eberle:
One more small thing: 

grafik

The bar with the dark background seems to have a fixed width. When I increase the width of the window, that width stays fixed and the right part of the bar is filled with a brighter colored background.

andreas-eberle:
@andreasb242: What's the status on this? Can you find the time to fix the small bugs, so we can get it into master?

jsettlers.main.swing/src/main/java/jsettlers/main/swing/lookandfeel/ui/ButtonUiStone.java
andreasb242: You have to ask Ben, I got the graphics with this names

andreasb242: Localize the header of the Dialog is no Problem. But this seems to be the Exception handler.
So to localize all messages, you have to localize all Exceptions like this: https://stackoverflow.com/questions/24988491/difference-between-e-getmessage-and-e-getlocalizedmessage

I generally prefer the way to handle the expected message, and provide a userfriendly string for common messages.

In all other cases, I display usually a general like, like "an error occurred, contact the developer bla bla", the error is: bla bla

Standard exceptions are never translated on windows with Oracle JDK, but always translated on Linux with OpenJDK (Connection Lost, File not found etc.)

@andreas-eberle What do you think?

andreas-eberle: @andreasb242: I think the map loading error is an error we have to expect (when you switch to a new version of JSettlers, you will probably end up with this error when you load a savegame / you might get it from original maps.) 

Therefore I think we should localize this error. I guess it should be possible to catch MapLoadingException and then display a special dialog heading and conent and add the message of the exception for debugging reasons. I think that would be kind of a middle way. What do you think?

andreas-eberle:
@andreasb242: Can you look at the strange headline label I posted above? I tried to find where I have to change something to fix it, but I don't find the right place.

andreasb242:
The error message is localized. I tried to skin the dialog, but it's more work than I thought, i first try to use the same background as on the main Panel, but it don't look good. Therefore I simple choose two different messages: Map loading failed and all other. On the other case the message from the Exception is still untranslated displayed.

andreas-eberle:
Thanks @andreasb242  and Ben! Looks much better now!

Issue #465, opened 2016-05-26
Work area selection wrong


derfuhrer2007:
I have selected exactly on the sawmill, however the selection was to the north. Did it again, same issue, it is way north than the sawmill area that I selected.

![screenshot 16](https://cloud.githubusercontent.com/assets/18621498/15585955/84b7fe7e-239f-11e6-8ad7-286c194b93d2.png)
[2016-05-26_23-22-52_Volcano_Mountain (3).zip](https://github.com/jsettlers/settlers-remake/files/285069/2016-05-26_23-22-52_Volcano_Mountain.3.zip)

michaelzangl:
Where was the work area before you clicked there?

andreas-eberle:
@michaelzangl: I think this is the same issue as #215. The problem is the calculated click position on high terrains.

derfuhrer2007:
Seems to be a problem when you give a target on a building. Observed this while moving troops.

@michaelzangl It was the default one

andreas-eberle:
Fixed with PR #672.

PR #672, opened 2017-08-15
MapCoordinateConverter

rpolzer:
can use height for view to map conversion

Rectangle selection of soldiers or pioneers up in the mountains
using the map "Settler Craters" had been impossible,
this is now working


rpolzer:
@andreas-eberle: this PR solves #215 and #465

andreas-eberle:
@rpolzer: Thanks for this fix!

PR #673, opened 2017-08-17
sound timing introduced

rpolzer:
movable sounds and action can be timed so they are synchronous


michaelzangl:
Nice.

For the switch/case/if: Can you use our styling guideline? We always add blocks, even for single line `if` and place every statement on a new line.

rpolzer:
Sound timing and worker action is closely related, so I pushed some more commits into this branch which improve optical details of the different workers.

rpolzer:
Here is a list of the changes added to this branch - all are minor optical and acoustical improvements:

lumberjack: hit the wood, bend down and raise up sequence
sawmiller: keep saw when taking a trunk, bend down and raise up sequence
stonecutter: door moved, bend down and raise up sequence
all miners: bend down and raise up sequence
iron melt: door moved, repeat melting without going home, bend down and raise up sequence, melting pot input moved, action: sound, smoke, molten metal
gold melt: repeat melting without going home, bend down and raise up sequence, action: sound, smoke, molten metal
charcoal burner: cough, bend down and raise up sequence
fisherman: uninterrupted animation, bend down and raise up sequence
crop farmer: scythe sound changed
pig farmer: sound changed, bend down and raise up sequence, pig feeding
mill: action duration same as and sound duration
butcher: sound, bend down and raise up sequence, pig stack moved into enclosure, meat stack moved out of the way
baker: bend down and raise up sequence, timing adjusted, bread position in the oven, smoke
donkey farmer: bend down and raise up sequence, feed and raise sequence changed

rpolzer:
I also muffled the birds a bit

rpolzer:
toolsmith and weaponsmith now make a bigger fire, as in the original game

jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/MapObjectDrawer.java
```diff
@@ -308,7 +309,8 @@ private void drawObject(int x, int y, IMapObject object, float color) {
 			break;
 
 		case SMOKE:
-			drawByProgress(x, y, 13, 42, progress, color);
+			int height = 30;
```
andreas-eberle: please make this to a private static final constant

jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/settlerimages/SettlerImageMap.java
```diff
@@ -346,6 +346,16 @@ public Image getImageForSettler(EMovableType movableType, EMovableAction action,
 	 */
 	private SettlerImageMapItem getMapItem(EMovableType movableType,
 			EMovableAction action, EMaterialType material, EDirection direction) {
+
+
+		if(movableType==null||action==null||material==null||direction==null){
```
andreas-eberle: use our formatter configuration to format the code. (spaces are missing between == and so on)

jsettlers.logic/src/main/java/jsettlers/logic/map/loading/original/data/EOriginalMapSettlersType.java
```diff
@@ -33,7 +33,7 @@
 	SAWMILLER(EMovableType.SAWMILLER),
 	FORESTER(EMovableType.FORESTER),
 	BOWMAN_L1(EMovableType.BOWMAN_L1),
-	MELTER(EMovableType.MELTER),
+	IRON_MELTER(EMovableType.MELTER),
```
andreas-eberle: is there also a GOLD_MELTER?

jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/MapObjectDrawer.java
```diff
@@ -430,62 +432,138 @@ public void draw(IMovable movable) {
...
+					soundNumber = 20;
+				}
+				break;
+			case HEALER:
+				if (delay > .8) {
+					soundNumber = 21;
+				}
+				break;
+			case GEOLOGIST: // TODO: should also check grid.getResourceAmountAt(x, y)
+				if (sound.random.nextInt(256) == 0) {
```
andreas-eberle: why are you doing the random here? In only one of 256 cases, you play the sound?

jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/MapObjectDrawer.java
rpolzer: "private" and "static" were flagged as error,
so I only added "final"

jsettlers.logic/src/main/java/jsettlers/logic/map/loading/original/data/EOriginalMapSettlersType.java
rpolzer: No - this was erroneous, I reverted this change

jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/MapObjectDrawer.java
rpolzer: else the geologists would shout far too often

andreas-eberle:
I'm trying to fix the tests. I found a bug in the savegame creation that appeared now in this PR but isn't caused by you... so I'm working to fix this.

jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/settlerimages/SettlerImageMap.java
```diff
@@ -346,6 +346,19 @@ public Image getImageForSettler(EMovableType movableType, EMovableAction action,
...
 	private SettlerImageMapItem getMapItem(EMovableType movableType,
 			EMovableAction action, EMaterialType material, EDirection direction) {
+
+
+		if(movableType == null
+				|| action == null
+				|| material == null
+				|| direction==null) {
+			int i = 0;
+			i++;
```
andreas-eberle: What is the idea of this? You declare a local variable i and increment it by one and then nothing happens with it?

jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/MapObjectDrawer.java
```diff
@@ -128,6 +128,8 @@
 
 	private static final float FLAG_ROOF_Z = 0.89f;
 
+	private static final int SMOKE_HEIGHT = 30;
```
andreas-eberle: @rpolzer: FYI: This is what I meant with private static final.

jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/settlerimages/SettlerImageMap.java
```diff
@@ -336,31 +330,21 @@ public Image getImageForSettler(EMovableType movableType, EMovableAction action,
...
 	 */
-	private SettlerImageMapItem getMapItem(EMovableType movableType,
-			EMovableAction action, EMaterialType material, EDirection direction) {
-
-
-		if(movableType == null
-				|| action == null
-				|| material == null
-				|| direction==null) {
+	private SettlerImageMapItem getMapItem(EMovableType movableType, EMovableAction action, EMaterialType material, EDirection direction) {
```
andreas-eberle: @rpolzer: I think your formatter might be configure to a line width of 80. We use a line width of 200 for our formatting.

andreas-eberle:
@rpolzer: The changes look good to me. However, we need to adjust the AI, since the changed walking paths and therefore timings of some buildings cause some problems to the AI making the "weaker" AIs actually stronger... I will try to get this sorted, but it will take some time. Nevertheless, many thanks for these cool changes!

codingberlin:
I looked at the AI tests.
They are green on my machine and thus the only reason why they break on Travis is this check:
https://github.com/jsettlers/settlers-remake/blob/master/jsettlers.logic/src/test/java/jsettlers/integration/ai/AiDifficultiesIT.java#L121

On my machine it takes maximum 480 ms to update the statistics. This check was introduced in the AI tests to prevent the implementation of too long running ai algorithms which would harm the frame rate. But it seems so that Travis is much slower than normal pcs. I would suggest to remove this check to erase the distraction by false alerts. If the AI would cause performance issues in the future there will come up issues from several users and than the issues can be fixed.

michaelzangl:
The hardcoded times in this test are already adjusted to the performance of a typical travis test server. The idea is to find the individual commit in which the AI became slow. Users tend to complain after months and we would have to search through hundreds of commits to find the issue.

I just re-triggered your build in case this was a false-alert (travis servers have some lag and are not really accurate for performance tests - but most of the times they do fine).

codingberlin:
@michaelzangl wit `git bisect` (binary search O(log ⁡ n)) you are very fast finding the performance issue causing commit.
Furthermore if the AI causes performance issues I would suggest to analyse its root cause and not to blame the last commit which was just the last straw which broke the camel's back.

@andreas-eberle 
Anyway its the performance of the AI and thus its logic (in this case the what and not the how) didn't need a patch.
First somehow travis must output the debug messages so that you can see which runtime breaks the test (ai logic or collecting ai statistic? median or maximum?)
Then the root cause can be searched and fixed.
You requested my help. I hope this helps you. For fighting with travis and analysing/fixing the performance I have not time cause of my little twins at the moment.

codingberlin:
It turned out that I forgot to rebuild the sources after switching in this branch and run the tests against master.
After rebuild two tests failed logically and not because of the runtime at my machine as well.
I found out that lack of solders is the problem for the hard AI in the battle with easy AI. I will investigate for a fix.

andreas-eberle:
Thanks @codingberlin for improving the AI to handle the changes!

andreas-eberle:
Thanks @rpolzer for implementing this!

Issue #677, opened 2017-09-29
Setting a material distribution to zero for all buildings causes bugs


tom-pratt:
Noticed a weird bug while developing the materials distribution menu on Android, and then was able to recreate the bug in the swing version.

If you set the distribution of any material to zero to all buildings for that material, then after that new buildings will not be built.

*Steps to reproduce*
- start a game
- go to distributions menu
- choose any material and set the distribution to all buildings to zero
- try and build a new building. the building will not be built

rpolzer:
solved - see pull request

PR #681, opened 2017-10-08
materials distribution all zero improved

rpolzer:
solves #677


jsettlers.logic/src/main/java/jsettlers/logic/map/grid/partition/manager/materials/requests/MaterialsForBuildingsRequestPriorityQueue.java
```diff
@@ -114,6 +114,9 @@ protected MaterialRequestObject getRequestForPriority(int priority) {
 	}
 
 	private int getRandomStartIndex() {
+		if (settings.drawRandomBuilding() == null){
```
andreas-eberle: @rpolzer: Please assign settings.drawRandomBuilding() to a variable and then use that variable later in line 120. This way you don't draw two (different) random buildings. Additionally, this should fix the replay test, as then the behavior should not be different then bevore.

rpolzer:
did I misunderstand you?

andreas-eberle:
@rpolzer: I had a look at it and found that you actually fixed two bugs. I separated the two fixes into two commits. I did this to make sure the one fix that is not expected to change the replay is separated from the second fix which is expected to change the replay (because the bug actually appeared in the replay) and therefore has changed replay reference files.

PR #686, opened 2017-12-17
Remove unconditional 'go.graphics.android' include from settings.gradle

ziarrek:
The settings.gradle file contains an unconditional include of 'go.graphics.android' which breaks the build process when trying to build the swing ui on a system with no Android installed (tested on Ubuntu 16.04).
Removing the line fixes the issue.
'go.graphics.android' is included anyway in line 25 if Android is present on the system.


michaelzangl:
Are you sure you based from a recent version? #644 should have solved this ~6 months ago

ziarrek:
Indeed. Must have messed sth up with updating my fork.
Sorry for the hassle, closing the PR!

PR #697, opened 2018-03-30
Update dependencies

tom-pratt:
Updated gradle and android dependencies.
Removed retro lambda.

The auto replay test failed in Travis. Not entirely sure what this is and haven't investigated!


andreas-eberle:
The AutoReplay files are different because a lambda is serialized and since the lambda implementation has changed, this changes the bytes of the savegame. So it is an expected change.

I wanted to regenerate the auto replay files by running "jsettlers.integration.replay.RegenerateAutoReplayITReferences" in jsettlers.tools. However, I get a `java.lang.ClassNotFoundException: org.junit.Assert` exception. @tom-pratt: Do you know what's wrong here?

tom-pratt:
Good idea. that's way tidier

tom-pratt:
Actually putting the variables in gradle.properties is also slightly annoying. Theyre all interpreted as strings so you have to use .toInteger() when you want to use it for things like targetSdk. Both ways seem kind of hacky to me and stop the IDE package management from working!

andreas-eberle:
I think we should then stick with the variables like we had it all the time.

andreas-eberle:
I updated gradle to version 4.6. This fixed the problems I had with IntelliJ not being able to run SwingManagedJSettlers. I also changed the dependencies of jsettlers.main.swing to compile so they are included in the jar.

PR #700, opened 2018-04-03
Improve BooleanMovingAverage

Simulant87:
Minor refactoring.
Renaming variable name to give it more expression of the held value.
fix typo in method name.
use Api function to fill array with default values

add test cases.


Simulant87:
This should not change any logic. I don't understand why the integration test is failing.

jsettlers.logic/src/test/java/jsettlers/algorithms/datastructures/BooleanMovingAverageTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+public class BooleanMovingAverageTest {
+
+
+    @Test
+    public void insertValue() {
```
andreas-eberle: I think this can be removed (empty method body)

PR #701, opened 2018-04-03
Use right primitive value type

Simulant87:
use long for Thread.sleep to prevent auto-boxing int to long.


Simulant87:
I found some more places of Thread.sleep() used with an int I want to imporve to directly use a long.

PR #702, opened 2018-04-05
Bugfix issue #638: replace the initial soldier request with only one swordman

Simulant87:
instead of requesting to get the building full of soldiers only one swordman is requested.
You are still able to remove the swordman request and add a pikeman manually.

This fixes https://github.com/jsettlers/settlers-remake/issues/638


paulwedeck:
Why did you rename this function ?

Simulant87:
the renaming was not necessary for the fix, but after reading the method body I thought the new name would be more expressive what the method does and makes it easier to understand and distinguish from the other methods.

Simulant87:
looks like my adjustment breaks the Replay integration test (makes sense) and the fight between easy AI and very easy AI becomes very even and does not finish in time.
How should this be handled to fix the tests?

PR #703, opened 2018-04-06
added command line parameter --maps 

### LINKED ISSUE #699: position from logs, maps, saves in ~/.jsettlers
I thought it would be great to have this game in linux-repositories.
So I created a PKGBUILD-script for Archlinux, to add jsettlers to the Arch User Repository. But there is a problem, which prevents me from creating a proper package for linux:

The files are put into the directory /opt/JSettlers/ (*.jars, maps) (readonly for normal user)
A script for launching the game is in /usr/bin/jsettlers:

```
#!/bin/sh
cd /opt/JSettlers
exec /usr/bin/java -jar /opt/JSettlers/JSettlers.jar "$@"
```

The Problem is that when starting jsettler from anywhere with the command
"jsettlers" the script changes the working directory to /opt/JSettlers because otherwise the maps are not found.

But /opt/JSettlers is readonly, so the logs and save -folders cannot be created.

Also if i remove the "cd /opt/JSettlers" and call "jsettlers" these folders (and maps) are created at the position where the working directory of the current commandline is. This is not perfect for linux because you should be able to start the game from everywhere and use always the same saves/maps/logs.

So my suggestion would be:
- look at the working dir for the maps-folder and add all maps provided by the game. (Remember they are readonly, so they are not editable with the MapCreator.)
- store new maps from the MapCreator in /home/<user>/.jsettlers (which already exists)
- load them also. (if maps exist in both use the one in /home/..)
- create logs always in /home/<user>/.jsettlers/logs
- saves should be also at /home/<user>/.jsettlers/save

What do you think about this?

aligator:
fixes #699

I found some time to add this option.
It is a simpler approach than I described in the issue, but it's enough for creating repo-packages. 

Now you can start it with --maps:
java -jar /usr/share/java/jsettlers/JSettlers.jar --maps=/usr/share/jsettlers/maps

And if you set the workingdir for example to ~/.jsettlers, packaging for linux is very easy.

If --maps is set, the workingdir is still used as additional maps-folder, because the MapCreator saves them there.

Without --maps it acts like before. 

As soon as this is pulled, I will add jsettlers to the Arch User Repositories. (maybe I will create packages for other distis, too.)

I have not tested it with Windows or Mac, yet.


jsettlers.logic/src/main/java/jsettlers/logic/map/loading/list/MapList.java
```diff
@@ -259,9 +259,14 @@ public MapList getMapList() {
 			return new MapList(getMapListers(), saveDirectory);
 		}
 
-		public void addResourcesDirectory(File resources) {
-			addMapDirectory(new DirectoryMapLister(new File(resources, "maps"), true));
-			saveDirectory = new DirectoryMapLister(new File(resources, "save"), true);
+		public void addResourcesDirectory(File savesFolder, File ... mapsFolders) {
+			for (File mapsFolder : mapsFolders) {
+				if (mapsFolder != null) {
```
michaelzangl: Don't use null array entries to indicate a missing entry. Just don't add it to the array then

jsettlers.logic/src/main/java/jsettlers/logic/map/loading/list/MapList.java
```diff
@@ -259,9 +259,14 @@ public MapList getMapList() {
 			return new MapList(getMapListers(), saveDirectory);
 		}
 
-		public void addResourcesDirectory(File resources) {
-			addMapDirectory(new DirectoryMapLister(new File(resources, "maps"), true));
-			saveDirectory = new DirectoryMapLister(new File(resources, "save"), true);
+		public void addResourcesDirectory(File savesFolder, File ... mapsFolders) {
```
michaelzangl: Using a List<File> is probably better here.

jsettlers.main.swing/src/main/java/jsettlers/main/swing/resources/SwingResourceLoader.java
```diff
@@ -91,7 +92,16 @@ public static void setupMapListFactory(String additionalMaps, File originalMapsF
 	}
 
 	private static void loadDefaultMapFolders(DefaultMapListFactory mapList) {
-		mapList.addResourcesDirectory(new File("."));
+		File mapsFolders[] = new File[2];
```
michaelzangl: Use an ArrayList<>();

jsettlers.main.swing/src/main/java/jsettlers/main/swing/resources/SwingResourceLoader.java
```diff
@@ -91,7 +92,16 @@ public static void setupMapListFactory(String additionalMaps, File originalMapsF
...
+
+		if (CommonConstants.MAPS_FOLDER != null) {
+			mapsFolders[0] = new File(CommonConstants.MAPS_FOLDER);
```
michaelzangl: Use ArrayList#add()

aligator:
I just found out that the --maps parameter already worked before. Don't know why I haven't tried it before I implemented it. My fault...
(But I haven't found any documentation of it anywhere, so it wasn't obvious.)

It is added inside of SwingResourceLoader:
```
// Setup map load paths
setupMapListFactory(options.getProperty("maps"), settlersFolderInfo.mapsFolder);
```
So this pull request has no sense.

But I'll add jSettlers to the Arch-repo, soon :-)

PR #705, opened 2018-04-08
Fix translations: ASK_DESTROY, game-quit-description, action_ASK_SET_WORK_AREA

Simulant87:


Simulant87:
maybe someone can translate the new label
action_ASK_SET_WORK_AREA = Set work area
into the missing languages.

jsettlers.graphics/src/main/resources/jsettlers/graphics/localization/labels_de.properties
```diff
@@ -127,12 +127,13 @@ resource_IRONORE = Eisen
...
 # actions
 action_BUILD = bauen
-action_DESTROY = einreißen
+action_ASK_DESTROY = einreißen
 action_MOVE_TO = gehe nach
 action_SET_WORK_AREA = Setze Arbeitsbereich
 action_START_WORKING = beginne zu arbeiten
 action_STOP_WORKING = höre auf zu arbeiten
 action_NEXT_OF_TYPE = Nächstes Gebäude dieses Typs
+action_ASK_SET_WORK_AREA = Arbeitsbereich setzen
```
andreas-eberle: Isn't this the same as action_SET_WORK_AREA in line 132?

Simulant87: action_SET_WORK_AREA is not working because of the missing "ASK". I was unsure if it might be used somewhere else but I think action_SET_WORK_AREA should be removed. We can also switch the woring in the german translation to match the old existing form.

PR #706, opened 2018-04-13
Fix minimap screen size

Simulant87:
The displayed screen size on the mini map was too big. So I fixed it to represent the actual part of the currently displayed map.

This is the mini map on the current master branch (too big)
onmaster

This is the mini map on my branch.
onbranch

I had to work with an offset = 50; of the calculation of size for the Y-coordinate. I think there is an underlying issue maybe in the MapCoordinateConverter#getMapForScreen on the calculation for the Y-coordinate that I just work around for the mini map calculation and that could be fixed in general. This would als help to calculate the correct center of the screen of the current view port. I also addapted this part but the incomming Y still seems to be wrong. The point is still a bit to high. So I am open for discussions and improvements.


jsettlers.common/src/main/java/jsettlers/common/map/shapes/MapRectangle.java
```diff
@@ -148,17 +145,17 @@ public boolean iterate(IBooleanCoordinateFunction function) {
...
+			return relativeY <= height && relativeX <= width;
 		}
 
 		@Override
 		public ShortPoint2D next() {
-			if (relativeY < height && width > 0) {
-				int x = getLineStartX(relativeY) + relativeX;
-				int y = getLineY(relativeY);
+			if (hasNext()) {
+				int x = minX + relativeX;
```
michaelzangl: Have a look at getLineStartX(), it's a bit more complicated than that ;-)

jsettlers.common/src/test/java/jsettlers/common/map/shapes/MapRectangleTest.java
```diff
@@ -0,0 +1,50 @@
+package jsettlers.common.map.shapes;
+
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+public class MapRectangleTest {
```
michaelzangl: Map rectangle is a rectangle on screen - so it is not a rectangle in map coordinates.

jsettlers.graphics/src/main/java/jsettlers/graphics/map/controls/original/panel/MainPanel.java
```diff
@@ -364,9 +364,9 @@ private void goBack() {
 
 	public void setMapViewport(MapRectangle screenArea, IGraphicsGrid grid) {
 		this.grid = grid;
-		displayCenter = new ShortPoint2D(screenArea.getLineStartX(screenArea.getLines() / 2)
-				+ screenArea.getLineLength() / 2, screenArea
-				.getLineY(screenArea.getLines() / 2));
+		short x = (short) (screenArea.getMinX() + (screenArea.getWidth() / 2));
```
michaelzangl: Use ints here - shorts for local variables don't save any space. ShortPoint2D has an int constructor for convenience.

jsettlers.graphics/src/main/java/jsettlers/graphics/map/minimap/Minimap.java
```diff
@@ -159,49 +159,47 @@ private void drawViewmark(GLDrawContext context) {
...
 		}
-		int lineStartX = mapViewport.getLineStartX(0);
-		int firstY = mapViewport.getLineY(0);
-		float minviewx = converter.getViewX(lineStartX, firstY, 0) * width;
-		float maxviewy = Math.min(converter.getViewY(lineStartX, firstY, 0), 1) * height;
-		float maxviewx = converter.getViewX(mapViewport.getLineEndX(0), firstY, 0)
-				* width;
-		int lastY = mapViewport.getLineY(mapViewport.getLines());
-		float minviewy = Math.max(converter.getViewY(lineStartX, lastY, 0), 0) * height;
+		int offset = 50;
```
michaelzangl: Why do you exactly need 50?

Simulant87: It was a trial and error. The result is a good fit. I did not find the underlying problem.

jsettlers.common/src/test/java/jsettlers/common/map/shapes/MapRectangleTest.java
Simulant87: That means the assumptions about the assertions are wrong? Should I remove the test?

jsettlers.common/src/main/java/jsettlers/common/map/shapes/MapRectangle.java
Simulant87: But the existing MapAreasTest did not fail with this implementation. Maybe it does not cover all the edge cases.

jsettlers.graphics/src/main/java/jsettlers/graphics/map/minimap/Minimap.java
Simulant87: MapContent uses a screen padding of 50, maybe this plays together?

jsettlers.common/src/test/java/jsettlers/common/map/shapes/MapRectangleTest.java
andreas-eberle: It might be the case that the test is wrong. It would be best if you can fix it.

andreas-eberle: It might be the case that the test is wrong. It would be best if you can fix it.

Simulant87: The tests are passing (also on master except for containsMax() because of an inconsistent usage of < or <= when checking the containment). So I don't understand what should be fixed.

jsettlers.graphics/src/main/java/jsettlers/graphics/map/minimap/Minimap.java
michaelzangl: Yes, this should be it. We should move it to a common constant so that if we change it one day, we won't miss the minimap.

Simulant87: Not sure if MapContent#SCREEN_PADDING = 50; is already the right constant and should be public.

andreas-eberle: Could you try to find out if its the right constant by chaning it locally to see what it changes and if it changes the minimap screen size accordingly?

Simulant87: I don't see a difference when increasing the value of SCREEN_PADDING or OVERDRAW_BOTTOM_PX. So I don't think there is a correlation.

andreas-eberle: Then leave it as a separate constant.

andreas-eberle:
@michaelzangl: Do you need any more changes? If not, please change your review to accepted.

michaelzangl:
No, thanks, that's it ;-)

andreas-eberle:
I found a small bug in the way the map viewed map area is drawn on the minimap (see image, marked in red). This happened after I zoomed out and then back in. The white line on the left side did not shrink as expected.

grafik

@Simulant87 @michaelzangl: Do you have an idea where this comes from / can fix this?

Simulant87:
@andreas-eberle at first I could reproduce this bug. But after switching to master (could not reproduce there) switching bach to the branch and doing a clean build of jsettlers.graphics I could no longer reproduce it. Can you reproduce it after a clean build?

andreas-eberle:
I was able to reproduce it the following way:
* start a new game where the player is on the lower left part of the map (don't know if this is required)
* zoom out and in again (staying in the lower left part of the map)
* move the screen area to the lower right corner. The bug should be growing while moving towards the lower right corner.

grafik

grafik

grafik

Simulant87:
ahh. I can only reproduce it in full screen mode. Then I also don*t have to zoom out and in before.
I will have a look.

andreas-eberle:
Thanks!

Simulant87:
@andreas-eberle fixed, extracted, and simplified the calculation.

andreas-eberle:
@Simulant87: Thanks!
@michaelzangl: Can you have a look at the new changes? It seems to work fine.

andreas-eberle:
@Simulant87: Thanks for this fix.

PR #708, opened 2018-04-14
select the right image for wine goods

Simulant87:
Wine uses currently the image of sulfur, see below:

goodsonmaster
templeonmaster

I changed the offset in the image file to pick the right image (bottle) for the wine:

goodsonbranch
templeonbranch


michaelzangl:
@Simulant87 This is because image indexes of gold/amazon/... editions are different.

If you want to, you can team up with @andreas-eberle. I just wrote a translation mechanism on the `enable_own_graphics` that can translate between the various versions. Andreas is currently figuring out the differing indexes.

My idea would be to always gold edition image indexes in the game and then translate to the other editions from there.

PR #709, opened 2018-04-14
Fix #506: Allow soldiers to be force-moved somewhere.

### LINKED ISSUE #506: Add a "force move" function for soldiers
Soldiers are attracted by attack targets quite intensively which is useful for unsupervised attacks, but prevents them from (efficient) retreat. Keeping the (sometimes) useful attraction and giving the possibility for retreat can be achieved with a force function for moving soldiers.

experienced with 0.3-alpha

michaelzangl:
@andreas-eberle Schau mal ob das auf Logik-Seite passt.


krichter722:
Nice, much-anticipated feature. I'm having trouble with soldiers moving to the enemy land no matter which how far away I bring them into my land (by clicking ~every second) with the nightly distribution 2018-06-08_21-13-14__24deb9a , so I'll delay the issue report and check again after this is merged.

andreas-eberle:
@krichter722: I fixed that particular problem (soldiers attack over huge distances) in #725.

Issue #663, opened 2017-07-27
Design fallback graphics for ferry and cargo ship


andreas-eberle:
This fallback graphic will be used when the user plays with the Amazon Demo which does not include the ship images.

jaschaio:
Would love to help out with this. Can you make it more actionable? I am not a java developer but designer and webdeveloper. What do you need? Which format? A transparent PNG in 8 different angles? Should they look like the original ships or something completely new? Could you supply me with the original images as I only have the Amazon Demo.

andreas-eberle:
Current status: @jaschaio is creating us ship images. I'll work on integrating them in the game.

andreas-eberle:
@jaschaio: What's the current status? I haven't heard from you in a quite a while now. I hope you are all fine! I'm looking forward to the ferry images! When we have them, we can create a new release with the ships :)

Issue #573, opened 2017-05-16
Implement detection of original Settlers installation type and choose image indexes accordingly


andreas-eberle:
### Situation

JSettlers relys on the original The Settlers III resource files. However, the content of these files is organized differently for each version of The Settlers III. This results in wrong images beeing shown depending on the original Settlers version you are using.

### Proposed fix

We need a system to detect the type of settlers installation that is available and choose a list of image indexes appropriately. This way, we can configure this index list for all versions correctly and can show the correct images for all versions.

rpolzer:
There could be another solution: as can be seen in the panel when selecting soldiers on the map, the panel shows the soldiers as they appear on the map, and not the "stone reliefs" which make trouble because of different indexes for different versions of the resource files.

Suggestion: for all statistics panels and for the tower population panel, we could use the images used on the map instead of the "stone reliefs".

michaelzangl:
Suggested fix:
(1) Use the gold edition indexes for all images
(2) Write a translation function that maps gold editon indexes -> standard edition / demo indexes.

mz8i:
@andreas-eberle what is left to do on this one? I noticed that #565 I reported some two years ago still persists and was thinking whether I could do some work on this issue

mz8i:
Upon some investigation, I understand that I basically need to use the DatFileMappingComparator to compare the Gold Edition GFX folder to my edition (I have a GOG Ultimate Collection edition) GFX folder and add the mapping to the source code. It is not yet entirely clear to me, though, what order I should run the comparison in. Any hints welcome!

PR #714, opened 2018-04-23
Allow using own graphics files and implement automated mapping of graphic files indexes

andreas-eberle:
This is required for #663.
Implements #573


andreas-eberle:
@jaschaio: No worries. I'm glad you have some progress on the images. 

There are three addtional things I need:

* To integrate the images into the game, we also need offsets for them. These offsets are needed to 1. get the center of the ships to where the ship is on the map and 2. put the parts of the ships together in the right place. Can you provide these offsets for your images? I guess you draw them in layers and thus that shouldn't be too much of a hassle, or is it? You can have a look at the images [I already added]( https://github.com/jsettlers/settlers-remake/tree/enable_own_graphics/jsettlers.common/raw_images_for_textures/cargo_ship). The .offset files are simple text files containing the x and y offset. Would be cool, if you can send me these offsets, so I don't have to find them by manually putting the images over each other and guessing them ;)

* I found out why there also is a complete boat image without a sail in the original graphics: It is used to display the boat while it is constructed. Can you also generate such images for every direction?

* We also need two very small boat images for the buttons in the dock where the player can select which type of boat he wants to be build. I can provide you the details for those via Discord.

No worries about the timing. We're glad to get these graphics. If you have any questions, just reach out to me on Discord.

jaschaio:
Hey Andreas, finally finished the last two images and sended it via discord.

**Offsets:**
This is indeed probably easier for me than for you as I have drawn them in layers in a photoshop file. Unfortunately I am unsure from which point exactly are those offsets calculated? Do you have photoshop and could "read them out" yourself? Or would an image like the following already help:

image
So for example measuring the offset from the uppermost left corner of the green image which is `0 0`, the pink would be `-1 -7` and the blue `51 25`. *(First the `Y` offset measured from the uppermost left corner of the left image and than the `X` offset)* Its missing the colored part of the sail but that should have the exact same measurements than the pole and as well the same offsets.

**Boats in Construction:**
Could you send me the reference images via discord? Should be easy to do as they are like I remember them.

**Buttons:**
Added to my todo 😅

PR #723, opened 2018-06-03
Extract new class that encapsulates logic about hashing of dat files.

ewaldbenes:


andreas-eberle:
@ewaldbenes: Thanks for the refactoring!

PR #731, opened 2018-06-14
draw stock front and back separately

rpolzer:
This produces a more realistic view of people going through the stock


andreas-eberle:
@rpolzer: When I construct a stock building, the building is already displayed, even though the construction has not even started.

andreas-eberle:
@rpolzer: Why did you make the positions marked with red blocked? Was this intentional?

grafik

(black positions are blocked, blue ones are protected)

andreas-eberle:
And sorry, I rebased your branch to get rid of all the merges. Please pull the branch again, if you want to add some more commits.

rpolzer:
The ranges marked in red are necessary to prevent stock walls and people to be drawn in the wrong order. The rear stock wall is neither exactly on y=-3 nor on y=-4, so I had to block two lines to prevent for example settlers showing their feet when standing behind the wall. The left front wall is especially critical, on 7 map positions either settlers or stock wall look ugly.

One of my merges was not necessary, but without doing it, I was not able to push - no idea why.

andreas-eberle:
@rpolzer: OK, thanks for adapting this. Just so you know: around all blocked partitions of a building, there must be protected positions. I just wrote this down in our wiki so it is documented for the future. https://github.com/jsettlers/settlers-remake/wiki/Implementation-Concepts---Important-Notes-for-Developers

PR #733, opened 2018-06-15
Use hash map to store settler images instead of 4-dim array

ewaldbenes:


jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/settlerimages/SettlerImageMap.java
```diff
@@ -295,51 +218,31 @@ public Image getImageForSettler(IMovable movable, float progress) {
...
-	 * @param direction
-	 * @param progress
-	 * @return The item of the map at the given position. Is not null.
+	 *
+	 * @param settlerImageFlavor@return The item of the map at the given position. Is not null.
 	 */
-	private SettlerImageMapItem getMapItem(EMovableType movableType, EMovableAction action, EMaterialType material, EDirection direction) {
-		SettlerImageMapItem item = this.map[movableType.ordinal()][action.ordinal()][material.ordinal][direction.ordinal];
+	private SettlerImageMapItem getMapItem(SettlerImageFlavor settlerImageFlavor) {
+		SettlerImageMapItem item = map.get(settlerImageFlavor);
```
andreas-eberle: Did you time the old and the new retrieval? I guess the performance of accessing a hash map vs. accessing a 4 dim array might be worse. How much is it for a larger number of calls (e.g. 1000 or so)?

jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/settlerimages/SettlerImageMap.java
```diff
@@ -38,40 +36,31 @@
...
-		this.directions = EDirection.VALUES.length;
-		this.map = new SettlerImageMapItem[this.types][this.actions][this.materials][this.directions];
+		imageProvider = ImageProvider.getInstance();
+	}
+
+	SettlerImageMap(ImageProvider imageProvider){
+		this.imageProvider = imageProvider;
+	}
 
+	public void loadFromMoveablesTextFile() {
```
andreas-eberle: typo "Movables" instead of "Mov[e]ables"

jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/settlerimages/SettlerImageMap.java
```diff
@@ -84,23 +73,18 @@ private SettlerImageMap() {
...
-		int[][][][] priorities = new int[this.types][this.actions][this.materials][this.directions];
-
-		// add pseudo entry.
-		addEntryToMap(priorities, null, null, null, null, DEFAULT_ITEM, -1);
-
-		readFromFile(file, priorities);
-	}
-
-	private void readFromFile(InputStream file, int[][][][] priorities) throws IOException {
+	private void readFromStream(InputStream file) throws IOException {
```
andreas-eberle: This method can be inlined. Both, the caller and this method, are now two-liners.

jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/settlerimages/SettlerImageMap.java
```diff
@@ -182,18 +168,18 @@ private EDirection parseDirection(final String directionString) {
 		return direction;
 	}
 
-	private int calculatePriority(EMovableType type, EMovableAction action, EMaterialType material, EDirection direction) {
+	private int calculatePriority(SettlerImageFlavor settlerImageFlavor) {
```
andreas-eberle: Move this method to SettlerImageFlavor, since it only accesses fields of the passed object.

jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/settlerimages/SettlerImageFlavor.java
```diff
@@ -0,0 +1,61 @@
...
+import jsettlers.common.movable.EMovableAction;
+import jsettlers.common.movable.EMovableType;
+import jsettlers.common.movable.IMovable;
+
+import java.util.Objects;
+
+public final class SettlerImageFlavor {
+
+    public static final SettlerImageFlavor NONE = new SettlerImageFlavor(null, null, null, null);
+    private final EMovableType type;
```
andreas-eberle: Use tabs as indents.

jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/ImageProvider.java
```diff
@@ -51,7 +53,8 @@
  *
  * @author michael
  */
-public final class ImageProvider {
+// TODO: removed final for testing. introduce it again when some supertype could be extracted
```
andreas-eberle: Remove the todo. The final isn't needed.

jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/settlerimages/SettlerImageMapItem.java
```diff
@@ -52,4 +52,31 @@ public int getStart() {
...
+				start == that.start &&
+				duration == that.duration;
+	}
+
+	@Override
+	public int hashCode() {
+		return Objects.hash(file, sequenceIndex, start, duration);
+	}
+
+    public int imageIndex(float progress) {
```
andreas-eberle: Use tabs as indent.

andreas-eberle:
I saw that the jsettlers.graphics:test wasn't executed by travis and added this.

jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/settlerimages/SettlerImageMap.java
michaelzangl: Actually, accessing a hash map should be faster than a 4 dimensional array (fewer memory accesses / cache misses).

jsettlers.graphics/src/main/java/jsettlers/graphics/map/controls/original/panel/selection/SelectionRow.java
```diff
@@ -48,10 +51,10 @@ public SelectionRow(EMovableType type, int count) {
 	@Override
 	public void drawAt(GLDrawContext gl) {
 		float width = getPosition().getWidth();
-		Image image =
-				SettlerImageMap.getInstance().getImageForSettler(type,
-						EMovableAction.NO_ACTION, EMaterialType.NO_MATERIAL,
-						EDirection.SOUTH_EAST, 0);
+		Image image = SettlerImageMap.getInstance().getImageForSettler(
+				new SettlerImageFlavor(type, EMovableAction.NO_ACTION, EMaterialType.NO_MATERIAL, EDirection.SOUTH_EAST),
```
michaelzangl: This is the main problem I see with your approach (and the reason why we did it the old way):

You are creating an object here that the JIT cannot (or at least in my tests with Android could not) inline. Mind that Android seems to be far worse at optimizing code than the normal Java JIT is.
This will make the young generation grow and thus create a lot of GC activity, since you might have thousands of settlers on one screen, especially when zooming out.
This is the reason why the drawing code attempts not to create any objects during painting phases.

We should test this and - if it is a problem - use a long instead of the SettlerImageFlavor (you can then create a static method that generates that unique long from the given parameters). Then use a library that supports long hash maps.

jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/settlerimages/SettlerImageMap.java
```diff
@@ -38,40 +36,31 @@
...
-	private static final SettlerImageMapItem DEFAULT_ITEM = new SettlerImageMapItem(10, 0, 0, 1);
+	static final SettlerImageMapItem DEFAULT_ITEM = new SettlerImageMapItem(10, 0, 0, 1);
 
 	private static SettlerImageMap instance;
 
-	private final ImageProvider imageProvider = ImageProvider.getInstance();
+	private final ImageProvider imageProvider;
 
-	private final SettlerImageMapItem[][][][] map;
+	final HashMap<SettlerImageFlavor, SettlerImageMapItem> map = new HashMap<>();
```
michaelzangl: private

jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/settlerimages/SettlerImageMapItem.java
```diff
@@ -52,4 +52,31 @@ public int getStart() {
...
+		SettlerImageMapItem that = (SettlerImageMapItem) o;
+		return file == that.file &&
+				sequenceIndex == that.sequenceIndex &&
+				start == that.start &&
+				duration == that.duration;
+	}
+
+	@Override
+	public int hashCode() {
+		return Objects.hash(file, sequenceIndex, start, duration);
```
michaelzangl: Same allocation problem here - use direct hash code computation instead to avoid allocating an array.

jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/settlerimages/SettlerImageMap.java
ewaldbenes: Access of hash map by key is 0(1) so the same as the existing index lookup.

Memory footprint is far less because the array was as big as all possible combinations. I measured over a million. By using a hash map it went down to a little more than 1000. The expressiveness and the lower memory footprint were the 2 major reason why I did this refactoring.

jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/settlerimages/SettlerImageMapItem.java
ewaldbenes: This object is used as a value in the hash map so no hashing is done. I don't know why I added hashCode and equals here ^^

jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/settlerimages/SettlerImageMap.java
michaelzangl: Actually the HashMap implementation of java has no guaranteed access of O(1) but a worst case access of O(n) instead ;-). But we don't care about worst case complexity here, we care about actual performance. Measuring this is hard - you cannot measure the GC delays you get in a few frames so easily. So you need to track speed and GC consumption with both the old and new methods in a reproducible setting (should be enough to test it on PC though, Android is even more complex for this).

jsettlers.graphics/src/main/java/jsettlers/graphics/map/controls/original/panel/selection/SelectionRow.java
ewaldbenes: You're right. I have overseen that fact. Maybe I can find a way to cache the SettlerImageFlavor instead of using a long since it would be the same approach. What do you think?

michaelzangl: The usual way to cache such objects is to put them in a hash table - You have a chicken and egg problem then ;-).

Generating the long is no big deal and should be pretty fast. A long -> Object hash map is probably the fastest you can get (in this case probably ~10 times faster than the way you are using now - at least that is my guess from my experience, might be more or less).

ewaldbenes: I've looked at this again. The SettlerImageFlavor can be created once in the constructor and the overhead is simply gone.

jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/settlerimages/SettlerImageMap.java
ewaldbenes: You said that the HashMap should be faster than the array. I'll look again at the hash function inside the flavor class and take care about object creation inside drawing methods to avoid GC pressure. Your two comments seem to contradict?

The array removal was merely a read- and maintainability change plus a huge memory gain.

jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/settlerimages/SettlerImageMap.java
ewaldbenes: It's used to assert contents in the map in the class test.

andreas-eberle:
@michaelzangl: Can you have a look at the updated PR?

jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/settlerimages/SettlerImageMap.java
```diff
@@ -132,14 +104,17 @@ private void addByLine(int[][][][] priorities, String line) {
...
-
 		final int fileIndex = Integer.parseInt(matcher.group(5));
 		final int sequence = Integer.parseInt(matcher.group(6));
 		final int start = Integer.parseInt(matcher.group(7));
 		final int duration = Integer.parseInt(matcher.group(8));
 
-		addEntryToMap(priorities, type, action, material, direction, new SettlerImageMapItem(fileIndex, sequence, start, duration), priority);
+		SettlerImageFlavor flavor = new SettlerImageFlavor(type, action, material, direction);
+		int priority = flavor.calculatePriority();
+		if(!priorities.containsKey(flavor) || priorities.get(flavor) < priority) {
```
michaelzangl: You can make this faster by using:
```
var value = priorities.get(flavor);
if (value != null || value ...) {}
```

jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/settlerimages/SettlerImageFlavor.java
```diff
@@ -0,0 +1,98 @@
...
+public final class SettlerImageFlavor {
+
+	public static final SettlerImageFlavor NONE = new SettlerImageFlavor(null, null, null, null);
+	private final EMovableType type;
+	private final EMovableAction action;
+	private final EMaterialType material;
+	private final EDirection direction;
+
+	public SettlerImageFlavor(EMovableType type, EMovableAction action, EMaterialType material, EDirection direction) {
+		this.type = type;
```
michaelzangl: You can ensure that the parameters are not null by adding Objects.requireNonNull() calls.

jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/settlerimages/SettlerImageFlavor.java
```diff
@@ -0,0 +1,98 @@
...
+		if (type != that.type) return false;
+		if (action != that.action) return false;
+		if (material != that.material) return false;
+		return direction == that.direction;
+	}
+
+	@Override
+	public int hashCode() {
+		int result = type != null ? type.hashCode() : 0;
+		result = 31 * result + (action != null ? action.hashCode() : 0);
```
michaelzangl: Same here - java is very efficient at testing for non-null and throwing an exception if a value is null.

jsettlers.graphics/src/main/java/jsettlers/graphics/map/draw/settlerimages/SettlerImageFlavor.java
ewaldbenes: The map contains one default entry that is defined by all parameters as null and comes from "movables.txt" file, entry "*,*,*,*=10, 0, 0, 1". * is converted to null. That's why I cannot prohibit null.

PR #734, opened 2018-06-21
replace nextDrawableX with x++

paulwedeck:


andreas-eberle:
@michaelzangl: Why did you introduce this in the first place? I guess there was an idea to it?

paulwedeck:
It has been introduced in bb15365a67b4a476391befa7e69de043c9fafac8

michaelzangl:
It was meant to skip tiles of which we definitely know that there is noting to draw there.

This would have made rendering a bit faster.

andreas-eberle:
"Nothing to draw" includes movables and objects?

andreas-eberle:
@paulwedeck: Please only merge master into your Pull Request if there are conflicts with your PR. Otherwise, I can easily merge it with Github directly before I merge it into master. This way, we can prevent a ton of merge commits.

michaelzangl:
@andreas-eberle Yes, no movables and no map objects.

paulwedeck:
@andreas-eberle I will try

PR #737, opened 2018-06-26
Dock shadow

rpolzer:
- dock drawing improved
- show correct ship shadow
- prevent ships from getting too close to obstacles


andreas-eberle:
Thanks for fixing this.

PR #739, opened 2018-07-02
Added missing menu titles

nptr:
* Added missing menu titles for DE & EN in the goods control panel.
* Fixed destroy-building confirmation dialog showing the actions name (which has no translation)


andreas-eberle:
Thanks for the fixes.

Issue #632, opened 2017-07-08
Wine-Grower wine-count


Silvergnom:
Greetings,

Just saw that the wine-grower doesnt show how many (yeah, countable in settlers 3 lmao) wine he already has.

![winer-grower wont tell me his secret](https://user-images.githubusercontent.com/29957146/27989093-db1831d4-6431-11e7-896d-cee7c5e0579b.png)

PR #740, opened 2018-07-03
Fixed offer stacks & building scripts

nptr:
* Added the missing offer stack to the winegrower. Resolves issue #632.
* Fixed the offer stack location of the baker
* Fixed baker being able to make bread with just water


nptr:
Elaboration on the baker fix:
The ```failjob``` branch of the material-availability-checks jumped to the ```A_wait``` action, being the end of the "take flour" act / begin of the "take water" act, effectively skipping "take flour". So despite missing flour, "take water" and "make bread" is executed, having the above mentioned effect of making bread just from water.

Furthermore, the drop location of the bread did not match the ```offerStack``` location. Also the location for the "offer stack not full" check was wrong, meaning the baker produced bread endless (wasting water and maybe flour).

With those problems fixed, new problems arose: failing unit/integration tests
* ```testStacksAreNotBlockedButProtected()```
  * the offerstack location is outside the buildings area (neither blocked nor protected), this is easily fixed by extending the buildings area or by moving the drop/offerstack location. I can do that.
* ```veryHardShouldProduceCertainAmountOfSoldiersWithin90Minutes()```
  * the fix changed the balancing. Maybe speeding up the wheat growth speed would fix that. I don't really want to decide that myself, change it, and hope it gets accepted. I'd like to have some input from the maintainers first.
* ```testReplay()```
  * now that the baker isn't overpowered anymore, the replay doesn't match its reference anymore. (At least i think the replayed state is compared to a static reference?) I'd like to have some input from the maintainers (@andreas-eberle , ...) here as well on how to progress and wether i am misunderstanding how this test works. Shall i just create a new reference once the balancing is fixed?

andreas-eberle:
@nptr: Thansk for working on this. To ensure that the answers to your questions will be findable in the future, I wrote them in our wiki. Here the links to the answers and reasons: 
* `testStacksAreNotBlockedButProtected`: Please mark the position as protected if it isn't too far away from the building. If it is, move the drop location to a better position. [details](https://github.com/jsettlers/settlers-remake/wiki/Implementation-Concepts---Important-Notes-for-Developers#stacks)
* `veryHardShouldProduceCertainAmountOfSoldiersWithin90Minutes`: Your change is expected to break that test since it reduces the amount of produced food. I think we will simply adapt the expected value here. If something else needs balancing, this should be done in a different PR. The AI seems to produce 834 soldiers instead of the expected 920. So this looks reasonable to me. I suggest changing the expected value to 830.
* `testReplay()`: [see here](https://github.com/jsettlers/settlers-remake/wiki/Implementation-Concepts---Important-Notes-for-Developers#autoreplayit)

nptr:
Here we go:
[+] marked stack position + NPC path as protected, was just 4 nearby points.
[+] increased the time limit for EASY vs. VERY_EASY because of the now delayed bread production
[+] reduced soldier requirement. Got 865 tho, not 834.

Yes, i might add a PR which buffs the AI again if necessary.

Issue #661, opened 2017-07-27
Allow to modify the selection of soldiers


michaelzangl:
This should be possible:
SHIFT => Add soldiers

Double-click or Alt-click: Select all of same type
CTRL-Click: Select all units in a given distance

TAB: Only select wounded

tom-pratt:
![screenshot_20170727-224648](https://user-images.githubusercontent.com/10612949/28691556-be5a0996-731d-11e7-849a-ad06cdebf1b2.png)

On Android we could have some of these options in the soldiers menu. It feels like you would have to make a normal selection of soldiers first for it to make sense, then the buttons are used to modify that selection.

A [+] button, press it then select more soldiers.
A button to select only the wounded out of the current selection.
Tap on the different soldier types to remove them from the current selection.

PR #746, opened 2018-07-13
Only select wounded in selection

PayDay1:
These changes allow you to filter your selection by wounded (key is TAB).
I also wanted to implement things like SHIFT while selecting to add to the selection, but I thought it is better to wait for #709 to be merged into master (because it already includes key modifiers for input events).

There is no UI for this in Android yet.

Should improve #661.


jsettlers.logic/src/main/java/jsettlers/input/GuiInterface.java
```diff
@@ -608,6 +614,21 @@ private boolean canSelectPlayer(byte playerIdOfSelected) {
 		return MatchConstants.ENABLE_ALL_PLAYER_SELECTION || playerIdOfSelected == playerId;
 	}
 
+	private void filterWounded() {
+		if(currentSelection.getSelectionType() != ESelectionType.SOLDIERS && currentSelection.getSelectionType() != ESelectionType.PEOPLE && currentSelection.getSelectionType() != ESelectionType.SPECIALISTS)
```
michaelzangl: Add a new Property to ESelectionType. Then use

```
if (currentSelection.getSelectionType().allowsFilterForWounded()) { ... }
```

jsettlers.logic/src/main/java/jsettlers/input/GuiInterface.java
```diff
@@ -608,6 +614,21 @@ private boolean canSelectPlayer(byte playerIdOfSelected) {
...
+			return;
+
+		final List<ISelectable> selected = new LinkedList<>();
+
+		for(ISelectable selectable : currentSelection) {
+			Movable movable = (Movable)selectable;
+			if(movable.getHealth() < movable.getMovableType().health)
```
michaelzangl: Add a new Method to ISelectable. Then use:
```
if (selectable.isWounded()) { ... }
```

Buildings always return false.

Then you can skip the test at the beginning of the method as well, since all selectables support the same interface.

PayDay1:
I implemented the changes that you guys have mentioned.
There is one questionable behaviour: If the selection only has unwounded objects in it, should it just do nothing or clear the selection (like it does now)?

michaelzangl:
I'm for clearing the selection.
So that you don't have to think about it when sending the wounded out of the battlefield. If there are no wounded, no soldiers will be sent.

andreas-eberle:
I would also say clear the selection.

PR #767, opened 2019-07-17
Many mostly graphics related optimizations

paulwedeck:
I reduced the number of shader to three:
- normal unified shader (now including texes and colors functionality) (GL(ES) 2.0)
- unified-array to bundle multiple drawings of a single quad at multiple locations/colors (GLES 3.0 /GL_EXT_draw_instanced
- unified-multi to bundle drawing of different objects at different locations (GLES32/ GL_EXT_geometry_shader4 &GL_ARB_uniform_buffer_object)

There no longer is any pre gl2 render path because there are barly any remaining users:
- for desktop gl (https://feedback.wildfiregames.com/report/opengl/feature/GL_VERSION)
- for android (https://developer.android.com/about/dashboards/index.html#OpenGL)

I rewrote the FogOfWar algorithm to better fit the Backgrounds needs and to respect to current game speed/pause status.

GeometryHandles are no replaced by *DrawHandles.
Most Textures and Geometries are bundled together in ManagedHandles hold by GLDrawContext that only create a single buffer and texture instead of one per SingleImage. Only parts of ManagedHandles can be used to draw via the unified-multi shader.


andreas-eberle:
Wow, this looks massive. I'll need some time to check this all out. Thanks already for your work!

paulwedeck:
419b2e2 fixes #769 for some characters (only that that I could extract from the labels_* files)

PR #771, opened 2019-07-20
Implement winning

### LINKED ISSUE #689: I can't win the game
Seems that even if I have conquered every last tower in single player I still haven't won the game.

tehKaiN:
fixes #689 

Stuff that works:
- Game checks if any player has lost by losing destroying all military buildings
- Game checks if any player has won by checking if all other players from other teams are defeated
- If player has won/lost his/her fog of war gets revealed and appropriate message displayed on top (vanilla S3 style)

Things I'm unsure implementation-wise
- I've made BuildingCounts public so that I can call it in WinLoseTracker and I'm unsure if this is acceptable
- I'm unsure if it's fine that WinLoseTracker generates new BuildingCounts each time it runs - I wanted to recount building, but perhaps there's a better way
- Everything else too. I'm new to the codebase and haven't dug to the bottom of it so I did some assumptions on how things work. I wouldn't be surprised if I did something in unclean way. ;)

Things I'm unsure of codestyle-wise:
- ~~I don't know if you accept ternary operator (`?:` condition)~~ I found it in InventoryPanel so I think it's not a problem
- I don't know if you accept early returns (check for inverse condition and breaking/continuing/returning on them to prevent too many indents)


tehKaiN:
Quoting @paulwedeck's message on discord:

> Why arent you calling disableFogOfWar directly in WinLoseTracker#timerEvent ?
> As far as I know Actions are there to notify any potential listeners of an UI interaction
> So atleast the Action stuff should be removed
> Maybe add something like getVisibleStatusPlayer in MainGrid.GraphicsGrid that returns the IPlayer from FogOfWars constructor
> If this is the just determined victor you call mainGrid.disableFogOfWar()

So I rewrote my code so that it disables FoW from `WinLoseTracker`'s `timerEvent()`

andreas-eberle:
@tehKaiN: Ternary and early returns are both allowed. In general, we aim to write readable code ;)

jsettlers.graphics/src/main/java/jsettlers/graphics/map/MapContent.java
```diff
@@ -374,6 +379,17 @@ private float messageAlpha(IMessage m) {
 				: Math.max(0, 1f - (float) age / IMessage.MESSAGE_TTL);
 	}
 
+	private void drawWinStateMsg(GLDrawContext gl) {
+		if(localPlayer.getWinState() == EWinState.UNDECIDED) {
```
andreas-eberle: Returning early like this is my preferred way! 👍

jsettlers.graphics/src/main/java/jsettlers/graphics/map/MapContent.java
```diff
@@ -374,6 +379,17 @@ private float messageAlpha(IMessage m) {
...
+			return;
+		}
+		Color color = (localPlayer.getWinState() == EWinState.WON ? Color.GREEN : Color.RED);
```
andreas-eberle: Why do you have the parenthesis around the ternary? They should not be needed.

andreas-eberle:
@tehKaiN: Thanks for your PR! Keep up the good quality!

I just found three small issues: 
* FoW wasn't disabled when a player lost 
* the players should not be allowed to do anything any more, if they lost the game.
* the win state was not persisted in the savegame. Therefore, after loading a savegame it was null.

I fixed these issues and will merge it when CI is green.

jsettlers.graphics/src/main/java/jsettlers/graphics/map/MapContent.java
tehKaiN: I don't know if I have an answer for that - that's the way I tend to write them. Feel free to remove it. ;) Or should I commit the fix?

tehKaiN:
Thanks for bug finding! I'll try to test my next changes more thoroughly.

